PR #1, opened 2021-04-28
Creating a Windows PowerShell installation script for the system agent.

phillipsj:
This is the start of the script for doing an system-agent install on Windows. The basics are there and it follows the same structure as the Bash version. Most of it has had some testing conducted and until we have an actual windows build it will be difficult to fully test.


phillipsj:
@Oats87 @luthermonson @sirredbeard @rosskirkpat

Issue #8, opened 2021-05-13
Add support for checking component status


brandond:
From @ibuildthecloud:

The agent should check component healthz or readyz status before indicating that the node install or upgrade is complete (ie updating the plan secret).
This would require checking:
* kubelet for agents
* apiserver and controller-manager for control-plane nodes
* etcd for etcd nodes

These endpoints usually require client cert authentication to access.

Oats87:
More information is needed around this. Technically, subsequent instructions can be defined that are healthchecks.

Oats87:
Synced with @ibuildthecloud this morning around this.

We are going to be building in Kubelet/Kubernetes-like healthz probing that is updated to the plan position (for local) and secret (for remote). 

The datastructure will be similar to:
```
	InitialDelaySeconds int32
	TimeoutSeconds int32
	PeriodSeconds int32
	SuccessThreshold int32
	FailureThreshold int32
```

in addition to a `URL`, `filepath` to a `cacert`, `client-cert`, and `key.

PR #13, opened 2021-05-21
Add healthchecking and optimize system-agent

Oats87:
https://github.com/rancher/system-agent/issues/8


pkg/applyinator/applyinator.go
```diff
@@ -28,6 +29,38 @@ type Applyinator struct {
 	imageUtil       *image.Utility
 }
 
+// CalculatedPlan is passed into Applyinator and is a Plan with checksum calculated
+type CalculatedPlan struct {
+	Plan     Plan
```
briandowns, line 34: Will this value be mutated by the applyinator?

pkg/applyinator/applyinator.go
```diff
@@ -43,22 +76,42 @@ func NewApplyinator(workDir string, preserveWorkDir bool, appliedPlanDir string,
 	}
 }
 
-func (a *Applyinator) Apply(ctx context.Context, anp types.AgentNodePlan) ([]byte, error) {
-	logrus.Infof("Applying plan with checksum %s", anp.Checksum)
+func CalculatePlan(rawPlanByteData []byte) (CalculatedPlan, error) {
```
briandowns, line 79: Can you remove the Hungarian notation? It provides an unnecessary redundancy.

pkg/k8splan/watcher.go
```diff
@@ -0,0 +1,198 @@
...
+				probePeriod = parsedPeriod
+			}
+		}
+		if secret == nil {
+			logrus.Debugf("[K8s] Secret was nil")
+			core.Secret().EnqueueAfter(w.connInfo.Namespace, w.connInfo.SecretName, probePeriod)
+			return secret, nil
+		}
+		secret = secret.DeepCopy()
+		logrus.Debugf("[K8s] Processing secret %s in namespace %s at generation %d with resource version %s", secret.Name, secret.Namespace, secret.Generation, secret.ResourceVersion)
```
briandowns, line 89: Would it be clearer if we changed `[K8s]` to `[system-agent]`?

pkg/k8splan/watcher.go
```diff
@@ -0,0 +1,198 @@
...
+				}
+			}
+
+			var wg sync.WaitGroup
+			var mu sync.Mutex
+
+			for probeName, probe := range cp.Plan.Probes {
+				wg.Add(1)
+
+				go func() {
```
briandowns, line 146:
```suggestion
				go func(wg *sync.WaitGroup) {
```

pkg/k8splan/watcher.go
```diff
@@ -0,0 +1,198 @@
...
+						probeStatus = prober.ProbeStatus{}
+					}
+					if err := prober.DoProbe(probe, &probeStatus, needsApplied); err != nil {
+						logrus.Errorf("error running probe %s", probeName)
+					}
+					mu.Lock()
+					logrus.Debugf("[K8s] (%s) writing probe status to map", probeName)
+					probeStatuses[probeName] = probeStatus
+					mu.Unlock()
+				}()
```
briandowns, line 164:
```suggestion
				}(&wg)
```

pkg/k8splan/watcher.go
Oats87: The reason for `[K8s]` is to help distinguish between `[K8s]` and `[local]` (source of plan data)

pkg/applyinator/applyinator.go
Oats87: Hungarian notation removed

pkg/applyinator/applyinator.go
Oats87: This value generally is not mutated by the applyinator

pkg/k8splan/watcher.go
```diff
@@ -0,0 +1,196 @@
...
+const (
+	appliedChecksumKey   = "applied-checksum"
+	appliedOutputKey     = "applied-output"
+	probeStatusesKey     = "probe-statuses"
+	probePeriodKey       = "probe-period"
+	planKey              = "plan"
+	enqueueAfterDuration = "5s"
+)
+
+func Watch(ctx context.Context, applyinator applyinator.Applyinator, connInfo config.ConnectionInfo) error {
```
briandowns, line 32: Does this need to return an error since it can't?

pkg/k8splan/watcher.go
```diff
@@ -0,0 +1,196 @@
...
+			if rawProbeStatusByteData, ok := secret.Data[probeStatusesKey]; ok {
+				if err := json.Unmarshal(rawProbeStatusByteData, &probeStatuses); err != nil {
+					logrus.Errorf("[K8s] error while parsing probe statuses: %v", err)
+					probeStatuses = make(map[string]prober.ProbeStatus, 0)
+				}
+			}
+			// calculate the checksum of the plan from the provided data
+			cp, err := applyinator.CalculatePlan(planData)
+			if err != nil {
+				return secret, err
```
briandowns, line 110: Should this return nil for the secret since there's an error?

Oats87: the convention is to return the object passed into the `OnChange`, so I don't believe we should be returning `nil`

pkg/k8splan/watcher.go
Oats87: error removed

briandowns (approved):

PR #41, opened 2021-08-02
Remove CA data if initial connection attempt fails and provide more context when unable to connect to Rancher

Oats87:
github.com/rancher/rancher/issues/33878

### DIFF
--- a/pkg/k8splan/watcher.go
+++ b/pkg/k8splan/watcher.go
@@ -8,15 +8,10 @@ import (
 	"strings"
 	"time"
 
-	"github.com/rancher/lasso/pkg/scheme"
-	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
-	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/runtime/schema"
-	"k8s.io/apimachinery/pkg/runtime/serializer"
-
 	"github.com/rancher/lasso/pkg/cache"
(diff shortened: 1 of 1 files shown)


pkg/k8splan/watcher.go
```diff
@@ -4,8 +4,15 @@ import (
...
 	"fmt"
+	"strings"
 	"time"
 
+	"github.com/rancher/lasso/pkg/scheme"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
+
```
luthermonson, line 15: nit

pkg/k8splan/watcher.go
```diff
@@ -51,15 +59,23 @@ func (w *watcher) start(ctx context.Context) {
 		panic(err)
 	}
 
+	if err := validateKC(ctx, kc); err != nil {
+		if strings.Contains(err.Error(), "x509: certificate signed by unknown authority") && len(kc.CAData) != 0 {
+			logrus.Debugf("Initial connection to Kubernetes cluster failed with error %v, removing CA data and trying again", err)
```
luthermonson, line 64: maybe this important enough to be an info log?

Oats87: moved to info log

luthermonson (approved):

PR #43, opened 2021-08-06
Don't apply if resource version is incorrect

Oats87:
https://github.com/rancher/rancher/issues/34027

### DIFF
--- a/pkg/k8splan/watcher.go
+++ b/pkg/k8splan/watcher.go
@@ -108,6 +108,10 @@ func (w *watcher) start(ctx context.Context) {
 		}
 		logrus.Debugf("[K8s] Processing secret %s in namespace %s at generation %d with resource version %s", secret.Name, secret.Namespace, secret.Generation, secret.ResourceVersion)
 		needsApplied := true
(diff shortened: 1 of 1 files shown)


pkg/k8splan/watcher.go
```diff
@@ -108,6 +108,10 @@ func (w *watcher) start(ctx context.Context) {
 		}
 		logrus.Debugf("[K8s] Processing secret %s in namespace %s at generation %d with resource version %s", secret.Name, secret.Namespace, secret.Generation, secret.ResourceVersion)
 		needsApplied := true
+		if w.lastAppliedResourceVersion > secret.ResourceVersion {
+			logrus.Debugf("received secret to process that was older than the last secret operated on. (%s vs %s)", secret.ResourceVersion, w.lastAppliedResourceVersion)
```
briandowns, line 112: Should we also add a similar line at INFO level but doesn't include the additional info?

Oats87: the logs will be spammed when doing something like this. we've opted to keep them at debug for this reason

briandowns (approved):

PR #46, opened 2021-08-18
Adding Windows builds and compiles.

phillipsj:
We need to take this step to start working on getting the system agent running on Windows.

Signed-off-by: Jamie Phillips <jamie.phillips@suse.com>

### DIFF
--- a/pkg/config/config_unix.go
+++ b/pkg/config/config_unix.go
@@ -1,0 +1,16 @@
+// +build !windows
+
+package config
+
+import (
+	"fmt"
+	"os"
+	"syscall"
+)
+
+func pathOwnedByRoot(fi os.FileInfo, path string) error {
+	if fi.Sys().(*syscall.Stat_t).Uid != 0 || fi.Sys().(*syscall.Stat_t).Gid != 0 {
+		return fmt.Errorf("file %s had was not owned by root:root", path)
+	}
+	return nil
+}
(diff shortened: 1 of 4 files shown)


Oats87 (approved):
LGTM

Oats87:
@briandowns can you also take a look at this?

pkg/config/config_unix.go
```diff
@@ -0,0 +1,16 @@
...
+package config
+
+import (
+	"fmt"
+	"os"
+	"syscall"
+)
+
+func pathOwnedByRoot(fi os.FileInfo, path string) error {
+	if fi.Sys().(*syscall.Stat_t).Uid != 0 || fi.Sys().(*syscall.Stat_t).Gid != 0 {
```
briandowns, line 12: Can you perform the type cast in the form of `if v, ok := whatever` and check the `ok` value?

scripts/version
```diff
@@ -6,7 +6,7 @@ fi
 
 COMMIT=$(git rev-parse --short HEAD)
 GIT_TAG=${DRONE_TAG:-$(git tag -l --contains HEAD | head -n 1)}
-
+CROSS=true
```
briandowns, line 9: Is this value used elsewhere?

pkg/config/config_unix.go
phillipsj: This was the existing code and I just moved it. Happy to make the changes.

scripts/version
phillipsj: That was already configured in the build script to enable cross compilation. I just used the existing hook to make it trigger.

pkg/config/config_unix.go
briandowns: Sorry. If you dont' mind. It'd be safer with the change.

briandowns (approved):

Issue #49, opened 2021-08-19
Agent attempts to take snapshots too quickly and for too long if s3 is misconfigured


davidnuzik:
**Summary:**
The rancher-system-agent service appears to take many snapshots in a very short timeframe (seconds) if s3 is misconfigured and uploading to the s3 bucket fails. Then after this brief time period, it will still keep trying, seemingly infinitely, over a longer time period to take snapshots and upload (about every 5 minutes). The details outlined below should provide more information and context.

**Environment:**
Rancher version: v2.6-head `2778ecaba` pulled at 8/19/21 12:51 pm Pacific
Rancher cluster type: single-node docker
Docker version: 20.10
 
Downstream cluster type: rke2
Downstream K8s version: v1.21.3-rc6+rke2r2

**Steps to Reproduce:**
1. In Rancher, create an 1-node RKE2 cluster and set up a mostly proper s3 config -- meaning correct keys, region, etc. Just don't supply a legit bucket name - enter a fake one. Tell Rancher to take the snapshot in the Rancher UI and it will. If you look at the rancher-system-agent logs this will fail but it does take snapshots locally. The agent does this many times for some reason in a very short period of time.
2. Watch the /var/lib/rancher/rke2/server/db/snapshots directory on the node. When you take the snapshot it takes many in a very short period.
3. If you simultaneously watch the rancher-system-agent systemd logs you'll see it loops very quickly over taking all these snapshots. It seems to back off and keep trying but more slowly after that point.
 
**Expected Result:**
I expected the rancher-system-agent to indeed retry taking a snapshot when it fails to upload to my s3 bucket, but not to retake snapshots so aggressively in such a short period/loop.

**Actual Result:**
The rancher-system-agent takes many snapshots in a very short period of time, then more slowly but seemingly infinitely over time.

Consider this directory listing of `/var/lib/rancher/rke2/server/db/snapshots` on the node:
```
root@dave-s3improper-pool1-gzrql:/var/lib/rancher/rke2/server/db/snapshots# ls -laht --time-style=full
total 159M
drwx------ 2 root root 4.0K 2021-08-19 20:20:34.348504465 +0000 .
-rw------- 1 root root  11M 2021-08-19 20:20:34.332504123 +0000 on-demand-dave-s3improper-pool1-gzrql-1629404434
-rw------- 1 root root  11M 2021-08-19 20:15:05.827968467 +0000 on-demand-dave-s3improper-pool1-gzrql-1629404105
[...]

PR #51, opened 2021-08-21
Add failure handling to system-agent

Oats87:
https://github.com/rancher/system-agent/issues/49

### DIFF
--- a/pkg/applyinator/applyinator.go
+++ b/pkg/applyinator/applyinator.go
@@ -173,7 +173,7 @@ func (a *Applyinator) Apply(ctx context.Context, cp CalculatedPlan) ([]byte, err
 	return gzOutput.Bytes(), nil
 }
 
-func (a *Applyinator) execute(ctx context.Context, prefix string, executionDir string, instruction Instruction) ([]byte, error) {
+func (a *Applyinator) execute(ctx context.Context, prefix, executionDir string, instruction Instruction) ([]byte, error) {
 	if instruction.Image == "" {
 		logrus.Infof("No image provided, creating empty working directory %s", executionDir)
 		if err := createDirectory(File{Directory: true, Path: executionDir}); err != nil {
--- a/pkg/applyinator/applyinator.go
+++ b/pkg/applyinator/applyinator.go
@@ -146,7 +146,8 @@ func (a *Applyinator) Apply(ctx context.Context, cp CalculatedPlan) ([]byte, err
 	for index, instruction := range cp.Plan.Instructions {
 		logrus.Debugf("Executing instruction %d for plan %s", index, cp.Checksum)
 		executionInstructionDir := filepath.Join(executionDir, cp.Checksum+"_"+strconv.Itoa(index))
-		output, err := a.execute(ctx, executionInstructionDir, instruction)
+		prefix := cp.Checksum + "_" + strconv.Itoa(index)
+		output, err := a.execute(ctx, prefix, executionInstructionDir, instruction)
 		if err != nil {
 			return nil, fmt.Errorf("error executing instruction %d: %v", index, err)
 		}
@@ -172,7 +173,7 @@ func (a *Applyinator) Apply(ctx context.Context, cp CalculatedPlan) ([]byte, err
 	return gzOutput.Bytes(), nil
 }
 
-func (a *Applyinator) execute(ctx context.Context, executionDir string, instruction Instruction) ([]byte, error) {
+func (a *Applyinator) execute(ctx context.Context, prefix string, executionDir string, instruction Instruction) ([]byte, error) {
 	if instruction.Image == "" {
 		logrus.Infof("No image provided, creating empty working directory %s", executionDir)
 		if err := createDirectory(File{Directory: true, Path: executionDir}); err != nil {
@@ -223,10 +224,10 @@ func (a *Applyinator) execute(ctx context.Context, executionDir string, instruct
 		writeLock sync.Mutex
 	)
 	eg.Go(func() error {
-		return streamLogs("[stdout]", &outputBuffer, stdout, &writeLock)
+		return streamLogs("["+prefix+":stdout]", &outputBuffer, stdout, &writeLock)
 	})
 	eg.Go(func() error {
-		return streamLogs("[stderr]", &outputBuffer, stderr, &writeLock)
+		return streamLogs("["+prefix+":stderr]", &outputBuffer, stderr, &writeLock)
 	})
 
 	if err := cmd.Start(); err != nil {
--- a/pkg/k8splan/watcher.go
+++ b/pkg/k8splan/watcher.go
@@ -28,11 +28,17 @@ import (
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/client-go/util/retry"
+	"k8s.io/client-go/util/workqueue"
 )
 
 const (
 	appliedChecksumKey   = "applied-checksum"
 	appliedOutputKey     = "applied-output"
+	failedChecksumKey    = "failed-checksum"
+	failedOutputKey      = "failed-output"
+	failureCountKey      = "failure-count"
+	successCountKey      = "success-count"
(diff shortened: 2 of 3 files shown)


pkg/applyinator/applyinator.go
```diff
@@ -172,7 +173,7 @@ func (a *Applyinator) Apply(ctx context.Context, cp CalculatedPlan) ([]byte, err
 	return gzOutput.Bytes(), nil
 }
 
-func (a *Applyinator) execute(ctx context.Context, executionDir string, instruction Instruction) ([]byte, error) {
+func (a *Applyinator) execute(ctx context.Context, prefix string, executionDir string, instruction Instruction) ([]byte, error) {
```
briandowns, line 176:
```suggestion
func (a *Applyinator) execute(ctx context.Context, prefix, executionDir string, instruction Instruction) ([]byte, error) {
```

pkg/applyinator/applyinator.go
```diff
@@ -146,7 +146,8 @@ func (a *Applyinator) Apply(ctx context.Context, cp CalculatedPlan) ([]byte, err
 	for index, instruction := range cp.Plan.Instructions {
 		logrus.Debugf("Executing instruction %d for plan %s", index, cp.Checksum)
 		executionInstructionDir := filepath.Join(executionDir, cp.Checksum+"_"+strconv.Itoa(index))
-		output, err := a.execute(ctx, executionInstructionDir, instruction)
+		prefix := cp.Checksum + "_" + strconv.Itoa(index)
```
briandowns, line 149: Do we know for certain `index` will always have an integer value in it?

Oats87: I believe so, `cp.Plan.Instructions` is a slice

pkg/k8splan/watcher.go
```diff
@@ -221,6 +307,27 @@ func (w *watcher) start(ctx context.Context) {
 	}
 }
 
+func (w *watcher) updateSecret(core corecontrollers.Interface, secret *v1.Secret) (*v1.Secret, error) {
+	var resultingSecret *v1.Secret
+	err := retry.OnError(retry.DefaultBackoff,
+		func(err error) bool {
+			if apierrors.IsConflict(err) {
```
briandowns, line 314: To be clear, we're not retrying if it's a conflict, but we will for other errors?

Oats87: Correct. This is a request that is originally from Darren, and will handle transient errors such as GOAWAY or server unavailable etc. 

The only time we will guaranteed not be able to update is if there is a conflict as the server will never take our version, so in that case we scrap what we're trying to do

pkg/k8splan/watcher.go
```diff
@@ -221,6 +307,27 @@ func (w *watcher) start(ctx context.Context) {
...
+				return false
+			}
+			return true
+		},
+		func() error {
+			var err error
+			resultingSecret, err = core.Secret().Update(secret)
+			return err
+		})
+	if err == nil {
```
briandowns, line 324: Can we refactor this so it looks for an error and actos accordily and returns nil for the resultingSecret and nil for the err at the end of the func?

pkg/k8splan/watcher.go
```diff
@@ -153,64 +192,111 @@ func (w *watcher) start(ctx context.Context) {
...
+				}
+			} else {
+				maxFailureThreshold = int(^uint(0) >> 1) // max int
+			}
+
+			logrus.Debugf("[K8s] Parsed max failure threshold value of %d", maxFailureThreshold)
+			wasFailedPlan := false
+			if rawFailureCount, ok := secret.Data[failureCountKey]; ok {
+				failureCount, err := strconv.Atoi(string(rawFailureCount))
+				if err == nil {
```
briandowns, line 211: Can we reverse this logic and check for the error and just log it if it exists? We an move the logic in this block out of this indentation.

pkg/k8splan/watcher.go
```diff
@@ -153,64 +192,111 @@ func (w *watcher) start(ctx context.Context) {
 				}
 			}
 
+			// Check to see if we've exceeded our failure count threshold
+			var maxFailureThreshold int
+			if rawMaxFailureThreshold, ok := secret.Data[maxFailuresKey]; ok {
+				// max failure threshold is defined. parse and compare
+				maxFailureThreshold, err = strconv.Atoi(string(rawMaxFailureThreshold))
+				if err != nil {
+					maxFailureThreshold = int(^uint(0) >> 1) // max int
```
briandowns, line 201: If no threshold is defined, is our default threshold 9,223,372,036,854,775,807?

pkg/k8splan/watcher.go
```diff
@@ -116,12 +159,8 @@ func (w *watcher) start(ctx context.Context) {
...
 		needsApplied := true
 		if toInt(w.lastAppliedResourceVersion) > toInt(secret.ResourceVersion) {
-			logrus.Debugf("received secret to process that was older than the last secret operated on. (%s vs %s)", secret.ResourceVersion, w.lastAppliedResourceVersion)
-			return secret, nil
-		}
-		if w.lastAppliedResourceVersion == secret.ResourceVersion {
-			logrus.Debugf("last applied resource version (%s) did not change. skipping apply.", w.lastAppliedResourceVersion)
-			needsApplied = false
+			logrus.Debugf("[K8s] received secret to process that was older than the last secret operated on. (%s vs %s)", secret.ResourceVersion, w.lastAppliedResourceVersion)
+			return secret, fmt.Errorf("secret received was too old")
```
briandowns, line 163:
```suggestion
			return secret, errors.New("secret received was too old")
```

pkg/k8splan/watcher.go
```diff
@@ -221,6 +307,27 @@ func (w *watcher) start(ctx context.Context) {
...
+		func() error {
+			var err error
+			resultingSecret, err = core.Secret().Update(secret)
+			return err
+		})
+	if err == nil {
+		logrus.Debugf("[K8s] updating lastAppliedResourceVersion to %s", resultingSecret.ResourceVersion)
+		w.lastAppliedResourceVersion = resultingSecret.ResourceVersion
+	}
+	return resultingSecret, err
```
briandowns, line 328: Is there value in returning this rather than the normal `nil` value?

Oats87: I'm assuming you're referring to the `resultingSecret` -- we need to return the `resultingSecret` back to the controller framework, which is why the logic is all wonky here.

pkg/k8splan/watcher.go
Oats87: I am working to retain the "standard" behavior which is to always return what is resulting from `core.Secret().Update()`, but we only want to update the last applied resource version in the event that the secret update was successful i.e. where `err == nil`

pkg/k8splan/watcher.go
Oats87: I didn't want to have the rest of this code hit in the event that the conversion was unsuccessful as we don't want to cause false positives of failure detection. It should basically never be that way unless someone manually manipulates the count value in the secret or corruption occurs.

pkg/k8splan/watcher.go
Oats87: Yes. I guess I could also just set it to `-1` and have logic to handle that case below, but by setting it to max int should provide more than ample amounts of attempts of a plan in my opinion.

pkg/k8splan/watcher.go
```diff
@@ -153,64 +193,114 @@ func (w *watcher) start(ctx context.Context) {
...
+				}
+			} else {
+				maxFailureThreshold = -1
+			}
+			wasFailedPlan := false
+			if rawFailureCount, ok := secret.Data[failureCountKey]; ok {
+				failureCount, err := strconv.Atoi(string(rawFailureCount))
+				if err != nil {
+					logrus.Errorf("[K8s] Error while parsing raw failure count: %v", err)
+					failureCount = 0
```
briandowns, line 214: Should we stop processing at this point?

PR #59, opened 2021-10-04
Parse CIDR on the fly in NO_PROXY env variable

thedadams:
The curl command doesn't parse CIDRs in the NO_PROXY environment
variable. This causes issues when trying to contact a private IP (in the
range 10.0.0.0/8 for example) because the proxy likely won't be able to
contact the private IP.

The change parses the NO_PROXY variable and determines if the IP address
that curl is attempting to contact is in the CIDR of any portion of the
NO_PROXY variable. If it is in a range, then the proxy will be ignored.

Issue:
https://github.com/rancher/rancher/issues/34908


install.sh
```diff
@@ -157,6 +157,107 @@ parse_args() {
     done
 }
 
+in_no_proxy() {
+    # Get just the host name/IP
+    ip_addr="${1#http://}"
+    ip_addr="${ip_addr#https://}"
+    ip_addr="${ip_addr%%/*}"
+    ip_addr="${ip_addr%%:*}"
```
thedadams, line 165: Update on 10/20/2021: added this one line.

PR #63, opened 2021-11-05
Fixes various Windows specific bugs discovered during test.

phillipsj:
File permissions on Windows don't behave the same as on Linux so those needed adjusting. 

Wharfie wasn't passing the OS information so the correct image wasn't being pulled for Windows and it was required to be bumped to address issues with extracting files from Windows images on Windows.

### DIFF
--- a/pkg/applyinator/file_unix.go
+++ b/pkg/applyinator/file_unix.go
@@ -1,0 +1,19 @@
+//go:build !windows
+// +build !windows
+
+package applyinator
+
+import (
+	"os"
+
+	"github.com/sirupsen/logrus"
+)
+
+// reconcileFilePermissions abstracts out the file permissions checks that only works on Linux.
+func reconcileFilePermissions(path string, uid int, gid int, perm os.FileMode) error {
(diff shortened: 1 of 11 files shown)


pkg/applyinator/file_unix.go
```diff
@@ -0,0 +1,22 @@
...
+	"github.com/sirupsen/logrus"
+)
+
+// reconcileFilePermissions abstracts out the file permissions checks that only works on Linux.
+func reconcileFilePermissions(path string, uid int, gid int, perm os.FileMode) error {
+	logrus.Debugf("reconciling file permissions for %s to %d:%d %d", path, uid, gid, perm)
+	if err := os.Chmod(path, perm); err != nil {
+		return err
+	}
+	if err := os.Chown(path, uid, gid); err != nil {
```
luthermonson, line 18: nit: you could just return os.Chown here instead of wrapping in an if

dweomer (approved):

rosskirkpat (approved):

PR #75, opened 2022-03-03
Only write applied plan contents if the plan actually changes

Oats87:
https://github.com/rancher/system-agent/issues/74


rosskirkpat (approved):
lgtm

PR #80, opened 2022-04-13
add periodic instruction cooldown 

Oats87:
https://github.com/rancher/rancher/issues/37296

This PR adds `max-retries` support and adds functionality to cooldown application of periodic instructions if they are failing.

### DIFF
--- a/pkg/k8splan/watcher.go
+++ b/pkg/k8splan/watcher.go
@@ -204,17 +204,19 @@ func (w *watcher) start(ctx context.Context) {
 
 			// Check to see if we've exceeded our failure count threshold
 			var maxFailureThreshold int
-			if rawMaxFailureThreshold, ok := secret.Data[maxFailuresKey]; ok {
+			if rawMaxFailureThreshold, ok := secret.Data[maxFailuresKey]; ok && len(rawMaxFailureThreshold) > 0 {
 				// max failure threshold is defined. parse and compare
 				maxFailureThreshold, err = strconv.Atoi(string(rawMaxFailureThreshold))
 				if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/k8splan/watcher.go
```diff
@@ -215,6 +216,21 @@ func (w *watcher) start(ctx context.Context) {
...
 			}
+
+			var maxRetries int
+			if rawMaxRetries, ok := secret.Data[maxRetriesKey]; ok {
+				// max failure threshold is defined. parse and compare
+				maxFailureThreshold, err = strconv.Atoi(string(rawMaxRetries))
+				if err != nil {
+					maxFailureThreshold = -1
+				} else {
+					logrus.Tracef("[K8s] Parsed max retries value of %d", maxRetries)
```
jakefhyde, line 227: Is this supposed to be the following? I don't see `maxFailureThreshold` anywhere else except the else block.
```suggestion
			var maxRetries int
			if rawMaxRetries, ok := secret.Data[maxRetriesKey]; ok {
				// max retries is defined. parse and compare
				maxRetries, err = strconv.Atoi(string(rawMaxRetries))
				if err != nil {
					maxRetries = -1
				} else {
					logrus.Tracef("[K8s] Parsed max retries value of %d", maxRetries)
```

jakefhyde (changes requested):

pkg/k8splan/watcher.go
Oats87: `maxFailures` is actually used on the rancher side. This logic can probably be cleaned up to only parse `max-failures` if `max-retries` is not set

thedadams (approved):

jakefhyde (approved):

jakefhyde (approved):

PR #82, opened 2022-04-26
Implemented uninstall script and added to install script

jakefhyde:
Implemented a simple uninstall script to remove the `CATTLE_AGENT_CONFIG_DIR`, `CATTLE_AGENT_VAR_DIR` the system-agent and uninstall script itself, and to stop and remove all system related resources. Verified working with system-agent `v0.2.5`. Additionally updated installer to allow installing uninstall script locally or from a different location (for parity with the binary installer).

Related Issue: https://github.com/rancher/rancher/issues/37283


jakefhyde:
@Oats87 Do you know if there are any other places I need to add the uninstall.sh script to the release artifacts?

snasovich:
@jakefhyde , how does uninstall for Windows fit with this? FYI: https://github.com/rancher/rancher/issues/37283#issuecomment-1109943140
cc: @phillipsj

jakefhyde:
@snasovich The associated PR for the windows agent can be found here: https://github.com/rancher/wins/pull/106

snasovich:
@jakefhyde , thank you. Actually noticed that PR a few seconds after posting my comment.

So on *nix we will have `uninstall.sh` per your PR present and instructions will be to just run this script.
What will be the situation in case of Windows? Will we have @phillipsj 's PowerShell script automatically copied too?

I also see that Windows implementation includes uninstalling RKE2 as part of their script while this one doesn't. I think we need to be consistent in both cases. @Oats87 , what are your thoughts on including RKE2 uninstall into this script?

jakefhyde:
I've implemented copying the uninstall script as part of the install script, we would likely need/want the same functionality on the windows side in the equivalent powershell script. Also, I think it's best that we uninstall rke2 separately.

Oats87:
Agreed. We should probably not uninstall RKE2 when running the system-agent uninstall script.

jakefhyde:
@phillipsj Based on [this comment](https://github.com/rancher/system-agent/pull/82#issuecomment-1112684944), we would like to pull the uninstall script in as part of the install script, as well as not uninstalling rke2 in order to have parity between the system-agent and it's windows counterpart.

phillipsj:
@jakefhyde okay, I will get those changes made.

phillipsj:
@jakefhyde @snasovich @Oats87 I removed the RKE2 uninstall step. Can I ask why we are downloading the uninstall script as part of the install? In addition, how will this work with Rancher? Currently the install script is pulled from a Rancher endpoint which will account for airgap, will we be adding an uninstall endpoint?

scripts/package-suc
```diff
@@ -8,6 +8,7 @@ cd $(dirname $0)/..
 mkdir -p dist/artifacts
 cp bin/rancher-system-agent dist/artifacts/rancher-system-agent${SUFFIX}
 cp install.sh dist/artifacts/install.sh
+cp uninstall.sh dist/artifacts/uninstall.sh
```
luthermonson, line 11: i think you need a line in the dockerfile too: https://github.com/rancher/system-agent/blob/main/package/Dockerfile.suc#L13 the drone job is just the docker plugin pointing to the .suc dockerfile: https://github.com/rancher/system-agent/blob/main/.drone.yml#L59 and this looks like a helper `make package-suc` for local dev

install.sh
```diff
@@ -316,7 +319,10 @@ setup_env() {
         fi
         BINARY_SOURCE=local
     else
-        BINARY_SOURCE=remote
+        if [ -z "${CATTLE_AGENT_BINARY_LOCAL_LOCATION}" ]; then
+            BINARY_SOURCE=remote
+        fi
+
```
thedadams, line 325: If `CATTLE_AGENT_BINARY_LOCAL` is not `true`, but `CATTLE_AGENT_BINARY_LOCAL_LOCATION` is set, then `BINARY_SOURCE` will be unset. Is this the desired outcome?

install.sh
```diff
@@ -337,6 +343,33 @@ setup_env() {
...
+        fi
+        UNINSTALL_SOURCE=local
+    else
+        UNINSTALL_SOURCE=remote
+        if [ -z "${CATTLE_AGENT_UNINSTALL_URL}" ] && [ -n "${CATTLE_AGENT_BINARY_BASE_URL}" ]; then
+            CATTLE_AGENT_UNINSTALL_URL="${CATTLE_AGENT_BINARY_BASE_URL}/uninstall.sh"
+        fi
+
+        if [ -z "${CATTLE_AGENT_UNINSTALL_URL}" ]; then
+            if [ $(curl --connect-timeout 60 --max-time 60 -s https://api.github.com/rate_limit | grep '"rate":' -A 4 | grep '"remaining":' | sed -E 's/.*"[^"]+": (.*),/\1/') = 0 ]; then
```
thedadams, line 358: Can we add an if-check on the `VERSION` so that if it is already set above, then we don't set it again here?

Also a nit: the version setting here and above is the same. Can this be extracted to its own function? It may not be because of uninstall versus install.

thedadams (changes requested):

install.sh
jakefhyde: I believe so. Admittedly I was confused by this initially, but this was what already existed.

thedadams: It looks like you are modifying what is existing.

jakefhyde: Fixed, not sure what I was thinking there

thedadams (approved):

annablender (approved):
More high level LGTM

Oats87 (approved):
LGTM, pending the Rancher change to include the uninstall script as an asset .

phillipsj (approved):

Issue #83, opened 2022-05-04
[bug] Windows does not support x509.SystemCertPool() in <go1.18


rosskirkpat:
The system cert pool in crypto/x509 library was re-enabled for Windows in go 1.18 but has not back-ported to 1.17 as of today. https://github.com/golang/go/commit/3544082f75fd3d2df7af237ed9aef3ddd499ab9c. This causes any secure probes that require a CA certificate to be loaded to fail in system-agent on Windows, which is implemented in rancher/wins.

The original upstream go issue https://github.com/golang/go/issues/16736 had a workaround https://github.com/golang/go/issues/16736#issuecomment-540373689 that I have implemented in system agent in a soon to be linked PR.

Rancher tracking issue: https://github.com/rancher/windows/issues/161

PR #84, opened 2022-05-04
add windows support for root CA cert stores

rosskirkpat:
### Summary

Move the get/create of a root CA cert store into OS dependent functions

Fixes https://github.com/rancher/system-agent/issues/83
Unblocks https://github.com/rancher/windows/issues/161

### Occurred changes and/or fixed issues

 Windows does not support x509.SystemCertPool() in <go1.18. This PR implements a workaround to load the System Cert Store on Windows nodes.

### Technical notes summary

This is adapted from the following workaround: https://github.com/golang/go/issues/16736#issuecomment-540373689

Error seen on Windows machines:
`Message : error loading system cert pool for probe (kubelet): crypto/x509: system root pool is not available on Windows`

Since <go1.18 on Windows cannot load the root CA cert store, this PR leverages syscalls to load the certificate context from the Root CA cert store and then create a new x509 cert pool that contains the Root CAs. This will likely cause a performance drop on Windows as each probe that requires a CACert will need to build it's own x509 cert pool with the root CAs.

Once system-agent is upgraded to go1.18+ the differing code paths should no longer be required.

### Areas or cases that should be tested

This PR cannot be tested before it is merged as far as I am aware.

### Areas which could experience regressions

These changes should only impact system agent functionality on Windows. The existing functionality for linux hosts has been shuffled into a new function but the underlying code is still the same. For windows, given that loading the CA certs from the root store is currently broken I do not see any potential for regression.

### Screenshot/Video
N/A

### DIFF
--- a/pkg/prober/prober.go
+++ b/pkg/prober/prober.go
@@ -2,6 +2,7 @@ package prober
 
 import (
 	"crypto/tls"
+	"crypto/x509"
 	"io/ioutil"
 	"net/http"
 	"net/url"
@@ -59,6 +60,7 @@ func DoProbe(probe Probe, probeStatus *ProbeStatus, initial bool) error {
 
 		caCertPool, err := GetSystemCertPool(probe.Name)
 		if err != nil || caCertPool == nil {
+			caCertPool = x509.NewCertPool()
 			logrus.Errorf("error loading system cert pool for probe (%s): %v", probe.Name, err)
 		}
 
--- a/pkg/prober/prober.go
+++ b/pkg/prober/prober.go
@@ -2,7 +2,6 @@ package prober
 
 import (
 	"crypto/tls"
-	"crypto/x509"
 	"io/ioutil"
 	"net/http"
 	"net/url"
@@ -58,13 +57,13 @@ func DoProbe(probe Probe, probeStatus *ProbeStatus, initial bool) error {
 			tlsConfig.Certificates = []tls.Certificate{clientCert}
 		}
 
-		caCertPool, err := x509.SystemCertPool()
+		caCertPool, err := GetSystemCertPool(probe.Name)
 		if err != nil {
-			caCertPool = x509.NewCertPool()
 			logrus.Errorf("error loading system cert pool for probe (%s): %v", probe.Name, err)
 		}
 
 		if probe.HTTPGetAction.CACert != "" {
+			logrus.Debugf("[DoProbe] adding CA certificate [%s] for probe (%s)", probe.HTTPGetAction.CACert, probe.Name)
 			caCert, err := ioutil.ReadFile(probe.HTTPGetAction.CACert)
 			if err != nil {
 				logrus.Errorf("error loading CA cert for probe (%s) %s: %v", probe.Name, probe.HTTPGetAction.CACert, err)
--- a/pkg/prober/prober.go
+++ b/pkg/prober/prober.go
@@ -58,7 +58,7 @@ func DoProbe(probe Probe, probeStatus *ProbeStatus, initial bool) error {
 		}
 
 		caCertPool, err := GetSystemCertPool(probe.Name)
-		if err != nil {
+		if err != nil || caCertPool == nil {
 			logrus.Errorf("error loading system cert pool for probe (%s): %v", probe.Name, err)
 		}
 
--- a/pkg/prober/prober_unix.go
+++ b/pkg/prober/prober_unix.go
@@ -5,6 +5,7 @@ package prober
 
 import (
 	"crypto/x509"
+	"fmt"
 
 	"github.com/sirupsen/logrus"
 )
--- a/pkg/prober/prober_unix.go
+++ b/pkg/prober/prober_unix.go
@@ -1,0 +1,21 @@
+//go:build !windows
+// +build !windows
+
+package prober
+
+import (
+	"crypto/x509"
+
+	"github.com/sirupsen/logrus"
+)
+
+// GetSystemCertPool returns a x509.CertPool that contains the
+// system certificates if they were present at runtime
+func GetSystemCertPool(probeName string) (*x509.CertPool, error) {
+	caCertPool, err := x509.SystemCertPool()
+	if err != nil {
+		caCertPool = x509.NewCertPool()
+		logrus.Errorf("[GetSystemCertPoolUnix] error loading system cert pool for probe (%s): %v", probeName, err)
+	}
+	return caCertPool, nil
+}
--- a/pkg/prober/prober_unix.go
+++ b/pkg/prober/prober_unix.go
@@ -10,7 +10,7 @@ import (
 )
 
 // GetSystemCertPool returns a x509.CertPool that contains the
-// system certificates if they were present at runtime
+// root CA certificates if they are present at runtime
 func GetSystemCertPool(probeName string) (*x509.CertPool, error) {
 	caCertPool, err := x509.SystemCertPool()
 	if err != nil {
--- a/pkg/prober/prober_unix.go
+++ b/pkg/prober/prober_unix.go
@@ -17,5 +17,8 @@ func GetSystemCertPool(probeName string) (*x509.CertPool, error) {
 		caCertPool = x509.NewCertPool()
 		logrus.Errorf("[GetSystemCertPoolUnix] error loading system cert pool for probe (%s): %v", probeName, err)
 	}
+	if caCertPool == nil {
+		return nil, fmt.Errorf("[GetSystemCertPoolWindows] x509 returned a nil certpool for probe (%s)", probeName)
+	}
 	return caCertPool, nil
 }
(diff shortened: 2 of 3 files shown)


phillipsj (approved):

pkg/prober/prober_windows.go
```diff
@@ -0,0 +1,75 @@
...
+				return nil, fmt.Errorf("[GetSystemCertPoolWindows] no certificate context was found for probe (%s)", probeName)
+			}
+		}
+		return nil, fmt.Errorf("[GetSystemCertPoolWindows] unable to enumerate certs in system cert store for probe (%s): %v", probeName, syscall.GetLastError())
+	}
+	if cert == nil {
+		return nil, fmt.Errorf("[GetSystemCertPoolWindows] certificate context returned from syscall is nil for probe (%s)", probeName)
+	}
+	// Copy the buf, since ParseCertificate does not create its own copy.
+	buf := (*[1 << 20]byte)(unsafe.Pointer(cert.EncodedCert))[:]
```
thedadams, line 55: Why do we need to copy the certificate? I understand that `x509.ParseCertificates` doesn't create its own copy, but why is that a problem?

It also looks like you are creating another copy with `buf2`. So we seem to have 3 copies?

In addition, why do we create an array here (with `[1 << 20]byte`) instead of creating a slice?

pkg/prober/prober_windows.go
```diff
@@ -0,0 +1,75 @@
...
+	buf2 := make([]byte, cert.Length)
+	copy(buf2, buf)
+	c, err := x509.ParseCertificate(buf2)
+	if err != nil {
+		return nil, fmt.Errorf("[GetSystemCertPoolWindows] unable to parse x509 certificate for probe (%s): %v", probeName, err)
+	}
+	certs = append(certs, c)
```
thedadams, line 62: It looks like `certs` will only ever have one certificate added to it. Why do we need `certs` here?

thedadams (changes requested):

pkg/prober/prober.go
```diff
@@ -58,13 +57,13 @@ func DoProbe(probe Probe, probeStatus *ProbeStatus, initial bool) error {
 			tlsConfig.Certificates = []tls.Certificate{clientCert}
 		}
 
-		caCertPool, err := x509.SystemCertPool()
+		caCertPool, err := GetSystemCertPool(probe.Name)
```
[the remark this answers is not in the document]
rosskirkpat: added a nil check

pkg/prober/prober_windows.go
rosskirkpat: I fixed this, I was missing the for loop that contains this code block

pkg/prober/prober_windows.go
```diff
@@ -0,0 +1,108 @@
...
+		buf := (*[1 << 20]byte)(unsafe.Pointer(certContext.EncodedCert))[:]
+
+		// buf2 is an array of bytes equal to the length of the certContext pointer
+		// which contains a certificate from the Root CA store
+		buf2 := make([]byte, certContext.Length)
+		copy(buf2, buf)
+
+		// validate the root CA certificate and return a x509.Certificate pointer
+		// that is appended into our array of x509 certificates
+		c, err := x509.ParseCertificate(buf2)
```
brandond, line 86: Can we perhaps just slice this to the correct length, instead of allocating a second slice of the correct length and then copying the bytes over?

```suggestion
		buf := (*[1 << 20]byte)(unsafe.Pointer(certContext.EncodedCert))[:certContext.Length]

		// validate the root CA certificate and return a x509.Certificate pointer
		// that is appended into our array of x509 certificates
		c, err := x509.ParseCertificate(buf)
```

brandond (changes requested):

pkg/prober/prober_windows.go
rosskirkpat: I wasn't sure how to do this but I knew there was a better way. thank you @brandond !

pkg/prober/prober_windows.go
rosskirkpat: This has been addressed with the latest commit.

luthermonson (approved):

rosskirkpat:
@snasovich If @Oats87 is not able to review this by EOD today as he is out of office, can I get your approval to merge once I have another round of approvals from @brandond and @thedadams ?

pkg/prober/prober_windows.go
```diff
@@ -0,0 +1,104 @@
...
+		}
+
+		// buf is a ~1048 kilobyte array that serves as a buffer holding the encoded value
+		// for each CA certificate in the Windows root CA store equal to the length of the certContext pointer
+		// which contains a certificate from the Root CA store
+		// we use a binary shift to create a ~1048 Kb buffer (slightly larger than 1 megabyte)
+		// [1 << 20]byte -> (1*2)^20 = 1048576 bytes
+		// maximum size of a Windows certificate store is 16 kilobytes and is not related to number of certificates
+		// (1048576*8)/4096 = 2048 (amount of 4096-bit certificates that can be stored)
+		buf := (*[1 << 20]byte)(unsafe.Pointer(certContext.EncodedCert))[:certContext.Length]
```
brandond, line 78: One more question - you said up above that
> 		// `CertEnumCertificatesInStore` will iterate through
>		// all available certContexts in the specified certificate store and
>		// return a single certContext containing the next certificate in the store

in that case, the certContext should only ever be 16kb, the max size of a single certificate entry, right? We don't need to over-allocate for the size of the entire store, since that's not what this is going to point at.

pkg/prober/prober.go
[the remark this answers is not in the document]
thedadams: I don't see a `nil`-check here.

It is possible for `GetSystemCertPool` to return `nil, error` in many cases, both for `unix` and `windows`. If this happens, we are only logging the error and continuing. Which means that `caCertPool` will be `nil` as we continue lower. This may cause later calls (like the on on line 71) to panic.

pkg/prober/prober_unix.go
```diff
@@ -0,0 +1,24 @@
...
+// root CA certificates if they are present at runtime
+func GetSystemCertPool(probeName string) (*x509.CertPool, error) {
+	caCertPool, err := x509.SystemCertPool()
+	if err != nil {
+		caCertPool = x509.NewCertPool()
+		logrus.Errorf("[GetSystemCertPoolUnix] error loading system cert pool for probe (%s): %v", probeName, err)
+	}
+	if caCertPool == nil {
+		return nil, fmt.Errorf("[GetSystemCertPoolWindows] x509 returned a nil certpool for probe (%s)", probeName)
+	}
```
thedadams, line 23: I think this can be removed.

thedadams (changes requested):

pkg/prober/prober_windows.go
rosskirkpat: I actually misinterpreted the syscall and that comment is not a proper description. It only fetches one cert at a time. I will update the comments I included to indicate that.

I am using the workaround referenced here, https://github.com/golang/go/issues/16736#issuecomment-540373689, which does not explain why they chose `(1*2)^20`. I assume it had to do with being divisible by 8 but it does make more sense to use 16kb. I'll get this change added

brandond: Hmm, I see some similar work in https://github.com/cloudfoundry/systemcerts/blob/master/root_windows.go, and I can't find anything in the MS API docs regarding the max size for the [_CERT_CONTEXT struct](https://docs.microsoft.com/en-us/windows/win32/api/wincrypt/ns-wincrypt-cert_context), so I guess 1024k is a fine worst-case value. You might update the comment to indicate that we're sizing for a single context (certificate) and not the whole store, though. That was a little confusing for me.

pkg/prober/prober_unix.go
rosskirkpat: I added the nil check inside each of the functions for linux/windows as it allows the function to fail faster for Windows and avoid a situation where a nil pointer causes a failure before we could check it in the main prober.go. Waiting to check until after the caCertPool is returned could result in a nil pointer being returned when we try to append the certs from the syscall to the caCertPool.

pkg/prober/prober_windows.go
brandond: Since we can't figure out exactly how large it'll be, maybe lets guard against making it too small? Otherwise, if we error on the short size, we'll crash when trying to access past the end of the array.

```suggestion
		// put this up at the top
		const maxEncodedCertLen := 1 << 20
		if certContext.Length > maxEncodedCertLen {
				return fmt.Errorf("invalid CertContext length %d, certContext.Length)
		}
		buf := (*[maxEncodedCertLen]byte)(unsafe.Pointer(certContext.EncodedCert))[:certContext.Length]
```

brandond (changes requested):

pkg/prober/prober_windows.go
rosskirkpat: agree w/ guarding against that situation, pushed the changes you suggested

brandond (approved):

thedadams (approved):

PR #85, opened 2022-05-06
Fix nil-pointer dereference on windows context

thedadams:
Issue: https://github.com/rancher/rancher/issues/37628

Passing certContext to the deferred function ensures that the value will
be taken at that point. The issue is that it is nil when the deferred
function is called.

This change will capture the variable so its real value is passed

### DIFF
--- a/pkg/prober/prober_windows.go
+++ b/pkg/prober/prober_windows.go
@@ -43,9 +43,11 @@ func GetSystemCertPool(probeName string) (*x509.CertPool, error) {
 	var certContext *syscall.CertContext
 
 	// ref for why flags value is 0: https://docs.microsoft.com/en-us/windows/win32/api/wincrypt/nf-wincrypt-certclosestore
-	defer func(store syscall.Handle, flags uint32) {
-		_ = syscall.CertCloseStore(store, flags)
-	}(certContext.Store, 0)
+	defer func() {
+		if certContext != nil {
+			_ = syscall.CertCloseStore(certContext.Store, 0)
+		}
+	}()
 
 	// this for loop will iterate through all available certificates in the specified certificate store
 	// and build an array of each x509.Certificate that is returned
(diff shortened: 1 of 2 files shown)


pkg/prober/prober_windows.go
```diff
@@ -43,9 +43,11 @@ func GetSystemCertPool(probeName string) (*x509.CertPool, error) {
 	var certContext *syscall.CertContext
 
 	// ref for why flags value is 0: https://docs.microsoft.com/en-us/windows/win32/api/wincrypt/nf-wincrypt-certclosestore
-	defer func(store syscall.Handle, flags uint32) {
-		_ = syscall.CertCloseStore(store, flags)
-	}(certContext.Store, 0)
+	defer func() {
```
rosskirkpat, line 46: The defer needs to be moved below the for loop on L94. I missed this on my initial PR, sorry!

rosskirkpat (changes requested):
move the original defer to L95 https://github.com/rancher/system-agent/blob/46fbba3a2040e7f2f4ef67802ee867dfe1cab9e8/pkg/prober/prober_windows.go#L95, after the for loop

pkg/prober/prober_windows.go
thedadams: Why does it need to be moved?

rosskirkpat: in addition to moving the defer, you would also do your original nil check after the for loop

luthermonson (approved):

rosskirkpat:
Were you able to validate this fixes the issue? You would need to pin this SA commit in a fork of wins and then compile wins and push the binary to a windows test machine.

luthermonson:
tested and the node came up

rosskirkpat (approved):
lgtm, thank you for testing!

PR #86, opened 2022-07-11
Bump to go1.18 and client-go 1.24, remove windows-specific x509 logic

rosskirkpat:
Removes the workaround implemented as part of fixing https://github.com/rancher/system-agent/issues/83

bumps to go1.18 and brings in 1.24 support


phillipsj (approved):

Oats87 (approved):
LGTM, but are we trying to ship this with 2.6.7?

rosskirkpat:
> LGTM, but are we trying to ship this with 2.6.7?

Yes, as we need to fix an issue with Windows workers for RKE2 encountered when using private CA for Rancher.

PR #90, opened 2022-08-16
Add opt bin dirs to PATH if /usr/local is readonly

ctrox:
Similar to #55, when installing on an OS that has a readonly `/usr` partition we ensure that the `rancher-system-agent` env knows about the bin path of rke2/k3s by extending the `PATH` in the systemd env file.

The main thing this fixes is listing etcd snapshots on such an OS as the `system-agent` will call the `rke2` binary for that.


vwbusguy:
You might consider doing this if /use/local is rw as well:

https://github.com/rancher/rancher/issues/40624#issuecomment-1652420486

PR #91, opened 2022-08-16
Add docker-publish-suc for s390x

rohitsakala:


.drone.yml
```diff
@@ -267,6 +267,28 @@ steps:
...
+  image: rancher/drone-images:docker-s390x
+  volumes:
+    - name: docker
+      path: /var/run/docker.sock
+  settings:
+    dockerfile: package/Dockerfile.suc
+    password:
+      from_secret: docker_password
+    repo: "rancher/system-agent"
+    tag: "${DRONE_TAG}-suc"
```
Oats87, line 280: This tag will clobber the existing `-suc` tag. 

Do we need to specify this as `-s390x-suc`? Notably, we don't support SUC upgrades on platforms beyond amd64 currently.

Oats87 (changes requested):

.drone.yml
rohitsakala: @Oats87 Fixed it. Added a manifest-suc pipeline.

Oats87:
@rohitsakala Have you tested this?

rohitsakala:
> @rohitsakala Have you tested this?

@Oats87 I ran docker build in an s390x system for Dockerfile.suc and it was built successfully. So I think it will work since the docker image is only alpine.

thehejik:
I successfully tested the image built out of this PR privided by Rohit `rohitsakala/system-agent-suc:s390x`.

To do so I had to replace the image in `system-agent-upgrader` Plan resource (in Rancher v2.6.8-rc4 on custom RKE2 cluster - Deployment -> `system-upgrade-controller` -> Related Resources tab -> Plan -> Edit)

I did the upgrade of custom RKE2 cluster twice from original 1.22 -> 1.23 -> 1.24. There is no trace anymore of related pending pod/jobs.

(ignore rancher/ ns typo in the screenshot)
plan-edit2

Oats87:
@rohitsakala would you mind running this manually using `drone exec`, publishing to a pipeline, and then providing proof that it 
a. publishes the correct images, and the manifest list
b. can be consumed by system-upgrade-controller in its published state, referencing the manifest-list?

I am just a little concerned that we will cause issues by changing from a manifest to a manifest-list.

rohitsakala:
> @rohitsakala would you mind running this manually using `drone exec`, publishing to a pipeline, and then providing proof that it a. publishes the correct images, and the manifest list b. can be consumed by system-upgrade-controller in its published state, referencing the manifest-list?
> 
> I am just a little concerned that we will cause issues by changing from a manifest to a manifest-list.

Hello @Oats87 ,

So, I ran `drone exec` from amd64 and s390x and checked the images in dockerhub. They were pushed correctly. 

S390x SUC Image - https://hub.docker.com/layers/rohitsakala/system-agent/dev-suc-s390x/images/sha256-587b80a28460b22c2b0bbdb30bc1ef2b6a4b817a1ac9890c6c05eaf43af30d3b?context=explore

AMD64 SUC Image - https://hub.docker.com/layers/rohitsakala/system-agent/dev-suc-amd64/images/sha256-ce6b3e880bb00ed0c07f21c60afdfcc53ae49812312d703a509ff6b870253f9e?context=explore

then I ran the manifest pipeline which result in this image - https://hub.docker.com/layers/rohitsakala/system-agent/dev-suc/images/sha256-ce6b3e880bb00ed0c07f21c60afdfcc53ae49812312d703a509ff6b870253f9e?context=explore


So, by this, we can say that the image pushing is done correctly. 

Now, for the second part, kubelet pulls the image from the docker hub. docker pulls the image according to the host architecture where the pod is run. So when I did `docker pull rohitsakala/system-agent:dev-suc`  it pulled the one used for amd64 and the same in s390x. 

For the second part, maybe @thehejik can prove that by replacing `rancher/system-agent:suc` with `rohitsakala/system-agent:dev-suc` and see if it works ?

Oats87 (approved):
Great, thank you for testing.
Please hold from merging until after Rancher 2.7.0 ships, in case we need to cut a last minute system agent.

PR #100, opened 2022-11-17
fix(run.sh): cleanup temp folder when script fail

FrankYang0529:
ref: https://github.com/harvester/harvester/issues/2821

When `run.sh` script fails, it leaves a temporary folder. Try to clean it up when the script exits.

### DIFF
--- a/package/suc/run.sh
+++ b/package/suc/run.sh
@@ -12,6 +12,9 @@ cleanup() {
     rm -rf "/host${TMPDIR}"
 }
 
+trap cleanup EXIT
+trap exit INT HUP TERM
+
 cp /opt/rancher-system-agent-suc/install.sh "/host${TMPDIR}"
 cp /opt/rancher-system-agent-suc/rancher-system-agent "/host${TMPDIR}"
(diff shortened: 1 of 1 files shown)


brandond (approved):

package/suc/run.sh
```diff
@@ -12,6 +12,8 @@ cleanup() {
     rm -rf "/host${TMPDIR}"
 }
 
+trap cleanup EXIT
```
Oats87, line 15: We may want to add a `trap exit INT HUP TERM` to catch signals as well, if we are enhancing the script this way.

Oats87 (commented):
Added a suggestion; depending on your thoughts on this, it LGTM if you don't want to add the change, let me know and I'll merge.

package/suc/run.sh
FrankYang0529: Do you mean update it like following?
```bash
trap cleanup EXIT
trap exit INT HUP TERM
```

Oats87 (approved):

PR #103, opened 2023-01-04
Update pipelines to support suc for arm64 and remove all arm support

sbstp:


sbstp:
The drone pipelines are currently failing because the `drone/git` image which is implicitly used by drone-ci no longer publishes an arm variant of the image. [This PR](https://github.com/drone/drone-git/pull/48) removed support for it.

sbstp:
Removed arm support as per @snasovich 's request.

Oats87 (approved):

PR #105, opened 2023-01-21
Add support for transactional-update based systems

rdoxenham:
I've introduced support for systems that utilise transactional-update, e.g. SLE Micro, openSUSE Leap Micro, and openSUSE MicroOS. On these systems the filesystem is either read-only, or is snapshot based, which means that the original install locations for CATTLE_AGENT_BIN_PREFIX and CATTLE_AGENT_VAR_DIR are not typically writeable, depending on whether transactional-update is being called.

This patch checks to see whether the system is using transactional-update and if so, adapts the install locations and/or calls transactional-update to make the appropriate changes.

This patch set supports both manually executing the script, i.e. a Rancher "custom" installation, and initiating the deployment via ignition/combustion (where transactional-update is called automatically).

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -80,6 +80,12 @@ check_target_ro() {
     test $? -ne 0
 }
 
+# check_rootfs_rw returns success if the root filesystem is read-write so we can check for transactional-update system
+check_rootfs_rw() {
+    touch /.rootfs-rw-test && rm -rf /.rootfs-rw-test
+    test $? -eq 0
+}
+
 # parse_args will inspect the argv for --server, --token, --controlplane, --etcd, and --worker, --label x=y, and --taint dead=beef:NoSchedule
 parse_args() {
     while [ $# -gt 0 ]; do
@@ -388,18 +394,30 @@ setup_env() {
         info "Using default agent configuration directory ${CATTLE_AGENT_CONFIG_DIR}"
     fi
 
+    # --- install to /var/lib/rancher/agent by default, except if we are running within transactional-update
+    # --- in which case we install into /etc/rancher/agent/var as /var is not mounted to the snapshot.
     if [ -z "${CATTLE_AGENT_VAR_DIR}" ]; then
-        CATTLE_AGENT_VAR_DIR=/var/lib/rancher/agent
-        info "Using default agent var directory ${CATTLE_AGENT_VAR_DIR}"
+        if [ -x /usr/sbin/transactional-update ] && check_rootfs_rw; then
+            CATTLE_AGENT_VAR_DIR=/etc/rancher/agent/var
+            info "Detected a transactional-update server, using ${CATTLE_AGENT_VAR_DIR} for agent var directory"
+        else
+            CATTLE_AGENT_VAR_DIR=/var/lib/rancher/agent
+            info "Using default agent var directory ${CATTLE_AGENT_VAR_DIR}"
+        fi
     fi
 
     # --- install to /usr/local by default, except if /usr/local is on a separate partition or is read-only
-    # --- in which case we go into /opt/rancher-system-agent.
+    # --- in which case we go into /opt/rancher-system-agent. If we are running within transactional-update
+    # --- we install to /usr as /usr/local and /opt are not mounted to the snapshot.
     if [ -z "${CATTLE_AGENT_BIN_PREFIX}" ]; then
         CATTLE_AGENT_BIN_PREFIX="/usr/local"
         if check_target_mountpoint || check_target_ro; then
             CATTLE_AGENT_BIN_PREFIX="/opt/rancher-system-agent"
             warn "/usr/local is read-only or a mount point; installing to ${CATTLE_AGENT_BIN_PREFIX}"
+        fi
+        if [ -x /usr/sbin/transactional-update ] && check_rootfs_rw; then
+            CATTLE_AGENT_BIN_PREFIX=/usr
+            warn "Detected transactional-update in progress; installing to ${CATTLE_AGENT_BIN_PREFIX}"
         fi
     fi
 


rdoxenham:
This helps address: https://github.com/rancher/rancher/issues/37539

install.sh
```diff
@@ -560,10 +576,16 @@ download_rancher_file() {
   local=$4
   local_location=$5
   source=$6
+  transactional=$7
+
+  TRANSACTIONAL_CMD=
+  if [ "${transactional}" = "true" ]; then
+      TRANSACTIONAL_CMD="transactional-update -d --no-selfupdate --continue --quiet run "
```
Oats87, line 583: There should be a way to override the`TRANSACTIONAL_CMD`, if a user needs to pass in specific or custom arguments to it.

Oats87: Furthermore, instead of calling it `TRANSACTIONAL_CMD`, we should probably rename this to `COMMAND_PREFIX` or something along those lines to allow for genericism. I do not look forward to having to rip out this environment variable later on, if we have to.

install.sh
```diff
@@ -577,7 +599,8 @@ download_rancher_file() {
           if [ "$(in_no_proxy "${url}")" = "0" ]; then
               noproxy="--noproxy '*'"
           fi
-          RESPONSE=$(curl $noproxy --connect-timeout 60 --max-time 300 --write-out "%{http_code}\n" ${CURL_BIN_CAFLAG} ${CURL_LOG} -fL "${url}" -o "${CATTLE_AGENT_BIN_PREFIX}/bin/${name}")
+          RESPONSE=$($TRANSACTIONAL_CMD curl $noproxy --connect-timeout 60 --max-time 300 --write-out "%{http_code}\n" ${CURL_BIN_CAFLAG} ${CURL_LOG} -fL "${url}" -o "${CATTLE_AGENT_BIN_PREFIX}/bin/${name}")
```
Oats87, line 602: I understand the `transactional-update` is supposed to allow for atomic application of changes to a node; however, this command tends to be very chatty in certain cases and can be run multiple times (if you look at the logic below). Is there a way we can mutate this logic to persist the resultant file with transactional-update without the "noise" of running transactional-update N times?

Possible solution for this would maybe be to stage the binary we are downloading to a temporary directory, then proceed to copy or move that binary from the temporary directory into the final destination (provided the transactional update can capture a `cp` or `mv`).

install.sh
```diff
@@ -845,9 +868,16 @@ do_install() {
     systemctl daemon-reload >/dev/null
     info "Enabling rancher-system-agent.service"
     systemctl enable rancher-system-agent
+
+    if [ ${TRANSACTIONAL_UPDATE} ] && [ -x /usr/sbin/transactional-update ]; then
+        info "Rebooting to envoke rancher-system-agent on transactional-update server."
+        systemctl reboot -f
```
Oats87, line 874: The brazen nature of this reboot is going to be a problem -- we run this install script from two potential places, either the standard shell for a user, or via a system-upgrade-controller job. In the prior case, I would say that we should provide some type of "opt out" or cancellation prompt (if interactive) to prevent the script from suddenly rebooting the user's machine, albeit they probably would expect it. 

On the secondary issue, we do not have granular control over when the system-upgrade-controller job runs. This means that on a Rancher upgrade, a user will potentially see their entire downstream cluster nodes reboot, with no rhyme or reason to it (we do not drain nodes for our SUC plan as of now, as this script is intended to be non-invasive and should not cause downtime). Level setting for this expectation of behavior will need to be done, before we can merge a change like this.

Oats87 (changes requested):
The support for `transactional-update` should really go through a formal design review before we start to cavalier the merging of functional changes that are this big.

install.sh
rdoxenham: Would you want it to be `COMMAND_PREFIX` or would `TRANSACTIONAL_COMMAND_PREFIX` work? Just thinking of longer term differentiation.

install.sh
rdoxenham: Yep, absolutely... with the default `transactional-update -d --no-selfupdate --continue --quiet run` command there's no chat to stdout - the only output is the desired "Successfully downloaded the...", and we can check the response code properly. We chain the changes together, i.e. downloading of the two files, into two operations that are stacked on-top of each other, that's why we use the `--continue` option, as otherwise multiple `transactional-update` commands would drop the previous run's changes. Here there's no need to stage the file and copy it in.

install.sh
rdoxenham: Yeah, I ignorantly thought that this script only executed once at installation time and rebooting a node wouldn't be a big deal. Hmm, on the first issue, do you think a "ctrl-c" type timeout would be preferable to an environment variable?

On the second, okay, I wasn't aware that we also run this script on a system-upgrade-controller job. We are only rebooting because of placement of `/usr/bin/rancher-system-agent` (and the uninstaller). I take it these files are also replaced during an upgrade with more recent versions if available? Or do they remain the same? If they remain the same, we could check for the existence of those files and skip the reboot, as they wouldn't be visible if they had just been installed (with transactional-update at least).

rdoxenham:
I've relooked at this and have come up with a much cleaner way of achieving the same. In my original commit I was assuming that the script would run once at installation time and that we really wanted the `rancher-system-agent` binaries to live in /usr somewhere. In this second revision-

1. We don't need to run `transactional-update` manually, we can redirect to /usr for `CATTLE_AGENT_BIN_PREFIX` if we're running within `transactional-update`, and /opt if we're running outside of `transactional-update` (checked with check_rootfs_rw and whether the command is available).

2. As with (1) we do the same for `CATTLE_AGENT_VAR_DIR`; if the rootfs is is writeable and `transactional-update` is available we write to /etc/rancher/agent/var, otherwise we're likely running normally, and we can use /var/lib/rancher/agent.

3. This approach does not require any rebooting, regardless of whether you're running through combustion or running the script manually. We're no longer trying to put things in /usr when `transactional-update` is running, hence no reboot is required to access these files normally.

Oats87 (approved):
On face value, this looks good to me, but we'll need to test it thoroughly before releasing it into the next version of `system-agent`. We'll need to test it both via a standard execution run, and via SUC.

jakefhyde (approved):

install.sh
```diff
@@ -388,19 +394,31 @@ setup_env() {
         info "Using default agent configuration directory ${CATTLE_AGENT_CONFIG_DIR}"
     fi
 
+    # --- install to /var/lib/rancher/agent by default, except if we are running within transactional-update
+    # --- in which case we install into /etc/rancher/agent/var as /var is not mounted to the snapshot.
     if [ -z "${CATTLE_AGENT_VAR_DIR}" ]; then
-        CATTLE_AGENT_VAR_DIR=/var/lib/rancher/agent
-        info "Using default agent var directory ${CATTLE_AGENT_VAR_DIR}"
+        if check_rootfs_rw && [ -x /usr/sbin/transactional-update ]; then
```
Oats87, line 400: Can we nest the check_rootfs_rw call under an if that checks for transactional update? I don’t think it makes sense for us to touch a file in / unless we know we have transactional update. That greatly reduces my regression concerns.

install.sh
```diff
@@ -388,19 +394,31 @@ setup_env() {
...
-    # --- in which case we go into /opt/rancher-system-agent.
+    # --- in which case we go into /opt/rancher-system-agent. If we are running within transactional-update
+    # --- we install to /usr as /usr/local and /opt are not mounted to the snapshot.
     if [ -z "${CATTLE_AGENT_BIN_PREFIX}" ]; then
         CATTLE_AGENT_BIN_PREFIX="/usr/local"
         if check_target_mountpoint || check_target_ro; then
             CATTLE_AGENT_BIN_PREFIX="/opt/rancher-system-agent"
             warn "/usr/local is read-only or a mount point; installing to ${CATTLE_AGENT_BIN_PREFIX}"
         fi
+        if check_rootfs_rw && [ -x /usr/sbin/transactional-update ]; then
```
Oats87, line 418: Same comment as above, it would be nice to not unnecessarily write to /.

Oats87 (changes requested):

install.sh
Oats87: I’m not in front of a computer but we may also just be able to reverse the order of the arguments in the existing if, as long as we stop evaluation once a failure occurs. I’m not in front of a computer able to test this atm though.

jakefhyde (approved):

Oats87 (approved):

PR #106, opened 2023-02-07
Use `kubectl version` to make sure the downloaded binary works

sbstp:
This fixes an issue where the `kubectl` being downloaded in the image is actually an XML document for a 404 error code. The fix is simply to transform `aarch64` to `arm64` when necessary. Additional protections are added to avoid this issue in the future:
* Use `curl -f` to fail on non 2xx status codes
* Run `kubectl version --client` to make sure the binary works as expected

### DIFF
--- a/package/Dockerfile.suc
+++ b/package/Dockerfile.suc
@@ -5,7 +5,7 @@ RUN ARCH=$(uname -m) && \
     if [ "$ARCH" = "x86_64" ]; then ARCH=amd64; fi && \
     curl -L -f -o /usr/bin/kubectl https://dl.k8s.io/release/v1.24.0/bin/linux/${ARCH}/kubectl && \
     chmod +x /usr/bin/kubectl
-RUN /usr/bin/kubectl version
+RUN /usr/bin/kubectl version --client
 
 ARG ALPINE=alpine:3.16.0
 FROM ${ALPINE}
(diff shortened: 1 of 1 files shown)


sbstp:
Fixes https://github.com/rancher/rancher/issues/40441 for arm64

jiaqiluo (approved):
LTGM.

package/Dockerfile.suc
```diff
@@ -3,8 +3,10 @@ FROM ${ALPINE} as kubectl
 RUN apk add -U curl
 RUN ARCH=$(uname -m) && \
     if [ "$ARCH" = "x86_64" ]; then ARCH=amd64; fi && \
-    curl -L -o /usr/bin/kubectl https://dl.k8s.io/release/v1.24.0/bin/linux/${ARCH}/kubectl && \
+    if [ "$ARCH" = "aarch64" ]; then ARCH=arm64; fi && \
+    curl -L -f -o /usr/bin/kubectl https://dl.k8s.io/release/v1.24.0/bin/linux/${ARCH}/kubectl && \
     chmod +x /usr/bin/kubectl
+RUN /usr/bin/kubectl version --client
```
HarrisonWAffel, line 9: Do we need this `RUN` step to check if the binary is working, or is this debug logging?

sbstp: It makes sure that the binary actually works for the arch it's being built on. It could catch bad files, downloading the wrong arch, etc.

HarrisonWAffel (approved):

package/Dockerfile.suc
snasovich: @sbstp , thank you for thinking about it and adding this "defensive" check here. Though since it's not exactly trivial why we do that, a comment would be nice in cases like this. (No need to go back and add it though)

PR #126, opened 2023-09-21
Evaluate applied-checksum after setting needApplied flag

anmazzotti:
We are running into some issues when the system-agent tries to re-apply remote plans that have been already applied.  

This happens whenever the system-agent is restarted or the entire host is rebooted.  

Some logic is already there to suppress re-execution based on secret resource version: 
```go
			if w.lastAppliedResourceVersion == secret.ResourceVersion && !wasFailedPlan {
				logrus.Debugf("[K8s] last applied resource version (%s) did not change. running probes, skipping apply.", w.lastAppliedResourceVersion)
				needsApplied = false
			}
```

However the `w.lastAppliedResourceVersion` is an in-memory value and a service restart will blank it again.
Therefore evaluating the `applied-checksum` is a more reliable way for remote plans.

### DIFF
--- a/examples/configuration/README.md
+++ b/examples/configuration/README.md
@@ -8,6 +8,7 @@ Create a file called `config.yaml` in `/etc/rancher/agent` with the contents lik
 workDirectory: /var/lib/rancher/agent/work
 localPlanDirectory: /var/lib/rancher/agent/plans
 remoteEnabled: true
+remoteSkipAlreadyApplied: false
 connectionInfoFile: /etc/rancher/agent/conninfo.yaml
 preserveWorkDirectory: true
 ```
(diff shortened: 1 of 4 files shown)


anmazzotti:
@davidcassany FYI

Oats87:
@anmazzotti

This is actually intentional behavior. The purpose for this is to force apply a plan if the system-agent is restarted or the node is rebooted.

I would be open to making this configurable; however, as we don't consider this behavior a bug, changing its default behavior is not acceptable

anmazzotti:
Thank you @Oats87 
I pushed an additional commit that introduces a flag to skip remote plans that have been already applied. 

Since the flag is false by default this should not impact the current logic.

PR #140, opened 2023-10-18
Fix: upgrade go in go.mod to 1.20 and container to latest golang version

matttrach:


matttrach:
this relates to SURE-7071

PR #150, opened 2023-12-11
Add interlocks to ensure operations are not interrupted

Oats87:
https://github.com/rancher/rancher/issues/43658

This PR adds interlocks to the system-agent which are used to indicate whether the install script and applyinator should continue.

The first interlock `restart-pending` indicates to the system-agent that a restart is pending per the install script, and disallows the applyinator from starting to apply a plan when a restart is pending.

The second interlock `applyinator-active` indicates to the install script that the applyinator is currently reconciling a plan and should not restart the system-agent.

The `restart-pending` interlock has a timeout feature built into the system-agent which allows the system-agent to ignore the pending restart after 5 minutes, which can be the case when the install script quits prematurely.

### DIFF
--- a/pkg/applyinator/applyinator.go
+++ b/pkg/applyinator/applyinator.go
@@ -30,6 +30,7 @@ type Applyinator struct {
 	workDir         string
 	preserveWorkDir bool
 	appliedPlanDir  string
+	interlockDir    string
 	imageUtil       *image.Utility
 }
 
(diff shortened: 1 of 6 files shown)


pkg/applyinator/applyinator.go
```diff
@@ -92,13 +93,17 @@ const defaultCommand = "/run.sh"
 const cattleAgentExecutionPwdEnvKey = "CATTLE_AGENT_EXECUTION_PWD"
 const cattleAgentAttemptKey = "CATTLE_AGENT_ATTEMPT_NUMBER"
 const planRetentionPolicyCount = 64
+const restartPendingTimeout = "5m"
```
HarrisonWAffel, line 96: It would be helpful to include a comment here that calls out the direct relation between this five minutes and the timeout in the `install.sh` script (via `APPLYINATOR_ACTIVE_WAIT_COUNT` and `sleep 5`)

HarrisonWAffel (changes requested):
a few small nits, otherwise makes sense to me as a way to block plan execution right before we restart

pkg/applyinator/applyinator.go
```diff
@@ -154,6 +159,66 @@ func (a *Applyinator) Apply(ctx context.Context, input ApplyInput) (ApplyOutput,
...
+						return output, fmt.Errorf("unable to write first-observed time to restart pending interlock file %s: %w", restartPendingInterlockFile, err)
+					}
+					return output, fmt.Errorf("restart is pending for system-agent")
+				} else {
+					// Parse the time out of the file and determine if we have passed our time threshold
+					t, err := time.Parse(time.UnixDate, string(fileContents))
+					if err != nil {
+						return output, fmt.Errorf("unable to parse first-observed time in restart pending interlock file %s: %w", restartPendingInterlockFile, err)
+					}
+					maxRestartPendingWait, err := time.ParseDuration(restartPendingTimeout)
```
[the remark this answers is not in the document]
Oats87: I just declared 5 minutes directly as a const of type duration

pkg/applyinator/applyinator.go
Oats87: They're actually opposite interlocks i.e. the install sh one is checking to see if the applyinator is applying a plan, whereas the restart-pending one stops application of a plan until the restart is done.

HarrisonWAffel: Ah okay, that makes sense

HarrisonWAffel (dismissed):

jakefhyde (dismissed):

HarrisonWAffel (approved):

jakefhyde (approved):

Issue #145, opened 2023-11-07
Retries succeed after getting an HTTP 401 error


lindhe:
I'm getting the following output when registering nodes in Rancher:

```console
$ curl -fL https://rancher.example.com/system-agent-install.sh | sudo  sh -s - --server https://rancher.example.com --label 'cattle.io/os=linux' --token f00barf00barf00barf00barf00barf00barf00barf00barf00bar --etcd --controlplane
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 30796    0 30796    0     0   252k      0 --:--:-- --:--:-- --:--:--  252k
[INFO]  Label: cattle.io/os=linux
[INFO]  Role requested: etcd
[INFO]  Role requested: controlplane
[INFO]  Using default agent configuration directory /etc/rancher/agent
[INFO]  Using default agent var directory /var/lib/rancher/agent
[INFO]  Successfully tested Rancher connection
[INFO]  Downloading rancher-system-agent binary from https://rancher.example.com/assets/rancher-system-agent-amd64
[INFO]  Successfully downloaded the rancher-system-agent binary.
[INFO]  Downloading rancher-system-agent-uninstall.sh script from https://rancher.example.com/assets/system-agent-uninstall.sh
[INFO]  Successfully downloaded the rancher-system-agent-uninstall.sh script.
[INFO]  Generating Cattle ID
[ERROR]  401 received while downloading Rancher connection information. Sleeping for 5 seconds and trying again
[ERROR]  401 received while downloading Rancher connection information. Sleeping for 5 seconds and trying again
[INFO]  Successfully downloaded Rancher connection information
[INFO]  systemd: Creating service file
[INFO]  Creating environment file /etc/systemd/system/rancher-system-agent.env
[INFO]  Enabling rancher-system-agent.service
Created symlink /etc/systemd/system/multi-user.target.wants/rancher-system-agent.service ÔåÆ /etc/systemd/system/rancher-system-agent.service.
[INFO]  Starting/restarting rancher-system-agent.service
```

The notable part is this:

```text
[ERROR]  401 received while downloading Rancher connection information. Sleeping for 5 seconds and trying again
[ERROR]  401 received while downloading Rancher connection information. Sleeping for 5 seconds and trying again
[INFO]  Successfully downloaded Rancher connection information
```

[...]

PR #151, opened 2024-01-04
Cleanup and exit if cannot download valid connection info.

vwbusguy:
https://github.com/rancher/rancher/issues/43991

This PR fixes an issue where if maximum retries are met attempting to download the rancher2_connection_info.json, the script continues with invalid data in that file causing the rancher-system-agent to fail to start.  This change cleans up the errant json file and exits the installer instead of continuing to setup a service doomed for failure.

Ref. SUSE Support case #01054437

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -774,7 +774,7 @@ retrieve_connection_info() {
                 ;;
             esac
         done
-        error "Max retries exceeded for downloading Rancher connection information."
+        error "Failed to download Rancher connection information in ${i} attempts"
         umask "${UMASK}"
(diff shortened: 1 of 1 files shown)


vwbusguy:
Addresses and possibly fixes #145

jakefhyde (changes requested):
Thank you for opening this PR. Overall the proposed change seems like a positive to me, and I don't see any immediate downsides.

jakefhyde (approved):
Approved, pending review by @Oats87

Oats87 (approved):

PR #197, opened 2024-09-24
Update Windows Files

HarrisonWAffel:
This PR closes the gap between how files are created on Linux and Windows. It also adds unit tests around file permissions, as well as updates CI to run these tests on both Linux and Windows. 

Note: This unit test does not compile properly on arm, and I can't determine why. Testing the CI on my fork repeatedly resulted in the below error. Because of this, I added a build tag to exclude this unit test on arm. I'm not aware of any differences between file creation and permissions between the two architectures, but if we feel strongly about including this test on arm I can continue to look into the error. But ultimately, for linux, this unit test is covering functionality was added to this repo ~3 years ago.

```  
fork/exec /tmp/go-build3394345592/b1239/applyinator.test: exec format error
FAIL	github.com/rancher/system-agent/pkg/applyinator	0.002s
```

### DIFF
--- a/pkg/applyinator/file_test.go
+++ b/pkg/applyinator/file_test.go
@@ -1,0 +1,300 @@
+//go:build !arm64
+
+package applyinator
+
+import (
+	"encoding/base64"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"reflect"
+	"runtime"
+	"testing"
+)
+
+func TestWriteContentToFile(t *testing.T) {
+	tempDir, err := os.MkdirTemp("", "")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer os.RemoveAll(tempDir)
+
+	getFile := func(path string, permissions string, content string) File {
+		return File{
+			Permissions: permissions,
+			Path:        filepath.Join(tempDir, path),
+			Content:     content,
+		}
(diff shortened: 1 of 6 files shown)


jakefhyde (dismissed):
Approved, weird about the arm test

pkg/applyinator/file_test.go
```diff
@@ -0,0 +1,300 @@
...
+		{
+			Path:    "test-no-perms",
+			Content: "hello world",
+
+			ExpectedPermissions: defaultFilePermissions,
+			ExpectedErr:         false,
+		},
+		{
+			Path:        "test-perms",
+			Permissions: "0600",
```
jiaqiluo, line 49: it seems that `defaultFilePermissions` is also 0600, could we change the value here, and in other cases, to a different one?

jiaqiluo (changes requested):
LGTM overall, with only one comment.

pkg/applyinator/file_test.go
HarrisonWAffel: I've updated the perms in `TestWriteContentToFile` to `0666`

jiaqiluo (approved):

jakefhyde (approved):

PR #201, opened 2024-10-10
Add ResetFailureCountOnServiceRestart

HarrisonWAffel:
Issue: https://github.com/rancher/rancher/issues/42458, RFC-13

This PR adds the `ResetFailureCountOnStartup` field to the `OneTimeInstruction` struct. If this field is set to true, during the [initial force application of the one-time instruction set ](https://github.com/rancher/system-agent/blob/7ad21ffbe9273d33d28fbf17249e949fdc6d8610/pkg/k8splan/watcher.go#L214-L218) the plans `failure-count` and `applied-checksum` fields will be reset. This ensures that the `max-failures` field is always respected, even if the plan had previously succeeded when executed by a prior instance of the system-agent. Due to the fact that the `failure-count` and `applied-checksum` fields are stored at the top level (and not as an attribute of the `plan` key) this will occur if any of the one time instructions set `ResetFailureCountOnStartup` to true.

### DIFF
--- a/pkg/applyinator/applyinator.go
+++ b/pkg/applyinator/applyinator.go
@@ -45,6 +45,9 @@ type Plan struct {
 	OneTimeInstructions  []OneTimeInstruction    `json:"instructions,omitempty"`
 	Probes               map[string]prober.Probe `json:"probes,omitempty"`
 	PeriodicInstructions []PeriodicInstruction   `json:"periodicInstructions,omitempty"`
+	// ResetFailureCountOnStartup denotes whether the system-agent should reset the failure count
+	// and applied-checksum for plans that are force applied each time the system-agent starts.
+	ResetFailureCountOnStartup bool `json:"resetFailureCountOnStartup,omitempty"`
 }
 
 type CommonInstruction struct {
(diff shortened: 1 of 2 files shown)


pkg/applyinator/applyinator.go
```diff
@@ -73,7 +73,8 @@ type PeriodicInstructionOutput struct {
 
 type OneTimeInstruction struct {
 	CommonInstruction
-	SaveOutput bool `json:"saveOutput,omitempty"`
+	SaveOutput                        bool `json:"saveOutput,omitempty"`
+	ResetFailureCountOnServiceRestart bool `json:"resetFailuresOnServiceRestart,omitempty"`
```
jakefhyde, line 77:
```suggestion
	ResetFailureCountOnStartup bool `json:"resetFailureCountOnStartup,omitempty"`
```
Name should match json, and I'd prefer `Startup` over `ServiceRestart` just due to the fact that it's shorter but has the same meaning.

jakefhyde (changes requested):
Smallest nit in the world

pkg/applyinator/applyinator.go
HarrisonWAffel: Updated, `r/r` pr is updated to use new json tag

jakefhyde (dismissed):

HarrisonWAffel:
After discussing this with @Oats87 this PR has been updated to move the `ResetFailureCountOnStartup` field from the `OneTimeInstruction` struct to the `Plan` struct

jakefhyde (approved):

Oats87 (approved):
lgtm

PR #202, opened 2024-10-11
Add GitHub App token generation and dispatch job for System Agent Upgrade workflow.

chiukapoor:
## Updates
- Add GitHub App token generation and dispatch job for System Agent Upgrade workflow.

## Issue
- https://github.com/rancher/rancher/issues/47179

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -271,9 +271,26 @@ jobs:
       - merge-suc
     permissions:
       contents: write # needed for creating the GH release
+      id-token: write
     steps:
+      - name: Read App Secrets
+        uses: rancher-eio/read-vault-secrets@main
+        with:
+          secrets: |
+            secret/data/github/repo/${{ github.repository }}/github/app-credentials appId | APP_ID ;
+            secret/data/github/repo/${{ github.repository }}/github/app-credentials privateKey | PRIVATE_KEY
+
+      - name: Create App Token
+        uses: actions/create-github-app-token@v1
+        id: app-token
+        with:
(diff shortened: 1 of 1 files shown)


snasovich (dismissed):

.github/workflows/release.yaml
```diff
@@ -296,9 +296,59 @@ jobs:
...
+    permissions:
+      contents: read
+      id-token: write
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    needs: github_release
+    strategy:
+      fail-fast: false
+      matrix:
+        target-branch: [release/v2.8, release/v2.9, main]
```
jakefhyde, line 330: Should this also include `release/v2.10`?

jakefhyde (changes requested):

.github/workflows/release.yaml
```diff
@@ -299,6 +316,45 @@ jobs:
...
+    permissions:
+      contents: read
+      id-token: write
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    needs: github_release
+    strategy:
+      fail-fast: false
+      matrix:
+        target-branch: [release/v2.8, release/v2.9, release/v2.10, main]
```
HarrisonWAffel, line 333: Are we actively updating 2.8 with new system-agents? If not, removing it might help keep the number of PRs raised to a minimum.

chiukapoor: Removed `release/v2.8`

HarrisonWAffel (approved):

jakefhyde (approved):

PR #228, opened 2025-04-02
Get UID/GID for current user in write file_test.go

jakefhyde:
This can just get merged whenever, it's just to enable these tests to run locally. I've confirmed they work both with `go test ./...` and `make test`.

### DIFF
--- a/pkg/applyinator/file_test.go
+++ b/pkg/applyinator/file_test.go
@@ -21,9 +21,11 @@ func TestWriteContentToFile(t *testing.T) {
 
 	getFile := func(path string, permissions string, content string) File {
 		return File{
-			Permissions: permissions,
-			Path:        filepath.Join(tempDir, path),
 			Content:     content,
+			UID:         -1,
+			GID:         -1,
+			Path:        filepath.Join(tempDir, path),
+			Permissions: permissions,
 		}
 	}
 
@@ -157,8 +159,10 @@ func TestCreateDirectory(t *testing.T) {
 	getFile := func(path string, permissions string) File {
 		return File{
 			Directory:   true,
-			Permissions: permissions,
+			UID:         -1,
+			GID:         -1,
 			Path:        filepath.Join(tempDir, path),
+			Permissions: permissions,
 		}
 	}
 


HarrisonWAffel (approved):

pkg/applyinator/file_test.go
jiaqiluo, line 1: Can we drop the build arg to allow it to run on arm64 now?

jiaqiluo (approved):

HarrisonWAffel:
Ah shoot, I forgot this test is shared between both platforms. On windows `u.Uid` is going to return [a security identifier](https://github.com/rancher/system-agent/actions/runs/14228101957/job/39872349106?pr=228) (e.g. `S-1-5-21-788787744-2135942908-530990601-500`) and not a simple integer, so atoi will fail. We should either do a GOOS check to conditionally pass the current uid and guid, or create platform specific test files (though they would be nearly identical).

jakefhyde:
> Ah shoot, I forgot this test is shared between both platforms. On windows `u.Uid` is going to return [a security identifier](https://github.com/rancher/system-agent/actions/runs/14228101957/job/39872349106?pr=228) (e.g. `S-1-5-21-788787744-2135942908-530990601-500`) and not a simple integer, so atoi will fail. We should either do a GOOS check to conditionally pass the current uid and guid, or create platform specific test files (though they would be nearly identical).

I'll opt for the former (for now).

jakefhyde:
@HarrisonWAffel actually from looking at the docs, it appears it does nothing if uid and gid are `-1` (means no change) so that might be a better option for the test. This will also be useful for any files the `system-agent` creates. Also FWIW it doesn't error when installed as a systemd service because they run as root (ID 0).

pkg/applyinator/file_test.go
jiaqiluo: Hi @jakefhyde, I think this question is still valid. Please let me know if I misunderstand anything.

jakefhyde: Let me try and see if it works

jakefhyde: We cannot. I'll create an issue to come back to this.

jakefhyde:
@HarrisonWAffel @jiaqiluo This should be ready for re-review now.

HarrisonWAffel (approved):

PR #231, opened 2025-05-21
Add delete for plan.File

kinarashah:
https://github.com/rancher/rancher/issues/50034


jakefhyde (approved):
Approved. FWIW If you feel strongly that it should be `delete` instead of `action` that's fine too.

HarrisonWAffel (approved):

PR #251, opened 2026-01-16
Add plan improvements proposal

alexander-demicev:
This PR adds a proposal for plan improvements and a template for future proposals

https://github.com/rancher/system-agent/issues/249

### DIFF
--- a/docs/proposals/20250115-plan-improvements.md
+++ b/docs/proposals/20250115-plan-improvements.md
@@ -1,0 +1,328 @@
+---
+title: System Agent Improved Plans
+authors:
+  - "@alexander-demicev"
+reviewers:
+  - TBD
+creation-date: 2026-01-15
+last-updated: 2026-01-16
+status: provisional
+replaces:
+superseded-by:
+---
+
+# System Agent Improved Plans
+
+## Table of Contents
+
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+
+- [System Agent Improved Plans](#system-agent-improved-plans)
+  - [Table of Contents](#table-of-contents)
+  - [Glossary](#glossary)
+  - [Summary](#summary)
+  - [Motivation](#motivation)
+    - [Current Limitations](#current-limitations)
+    - [Goals](#goals)
+    - [Non-Goals/Future Work](#non-goalsfuture-work)
+  - [Proposal](#proposal)
+    - [User Stories](#user-stories)
+    - [Implementation Details/Notes/Constraints](#implementation-detailsnotesconstraints)
+      - [NodePlan CRD](#nodeplan-crd)
+      - [File Locking and Multi-NodePlan Coordination](#file-locking-and-multi-nodeplan-coordination)
+      - [Preventing Spec Changes During Execution](#preventing-spec-changes-during-execution)
+      - [Content References for Large Files](#content-references-for-large-files)
+      - [OCI-Packaged Plans with Signature Verification](#oci-packaged-plans-with-signature-verification)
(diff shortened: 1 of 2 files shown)


docs/proposals/20250115-plan-improvements.md
```diff
@@ -0,0 +1,328 @@
...
+### Current Limitations
+
+1. **Extensibility**: Adding features like custom probes or retry strategies requires big changes throughout the codebase
+2. **Observability**: Limited structured feedback for operations
+3. **Multi-party coordination**: Only Rancher can currently manage plans; other parties have no safe coordination mechanism
+4. **Secret size limits**: Plan content can exceed 1MiB Secret limit due to inline base64-encoded files
+5. **Schema validation**: No validation of plan structure until runtime execution
+6. **Security**: No verification of plan content before execution
+
+### Goals
```
HarrisonWAffel: I also recall previous discussions around scalability and the system agent. Right now we have a 1-to-1 mapping of nodes and their plan secrets (or their CRDs in this approach), which at scale can cause etcd pressure. Is there any way this new approach could be used to reduce the number of plans needed for a given cluster? This might be a more long term goal, or may have changed, it's been a while since it was brought up

docs/proposals/20250115-plan-improvements.md
```diff
@@ -0,0 +1,328 @@
...
+  
+  locking:
+    enabled: true  # Default
+  
+  execution:
+    timeout: "30m"
+
+status:
+  observedGeneration: 5
+  phase: Pending|Executing|Applied|Failed|Cancelled
```
jakefhyde: One thing to note here is that the system-agent will reconcile plans on startup. This could be configurable, but this is generally a good idea as it ensures the config is applied.

docs/proposals/20250115-plan-improvements.md
```diff
@@ -0,0 +1,328 @@
...
+      probe: { httpGet: {...} }
+      required: true
+  
+  locking:
+    enabled: true  # Default
+  
+  execution:
+    timeout: "30m"
+
+status:
```
jakefhyde: I would still likely hash the plan somewhere

docs/proposals/20250115-plan-improvements.md
```diff
@@ -0,0 +1,328 @@
...
+
+#### File Locking and Multi-NodePlan Coordination
+
+Multiple NodePlans can exist for the same node (created by Rancher, or other parties). To ensure only one plan executes at a time, the agent uses file-based locking.
+
+**How it works**:
+- Agent watches ALL NodePlans with label `cattle.io/node: <nodeName>`
+- Before executing, reconciler attempts to acquire a file lock at `/var/lib/rancher/agent/plan.lock`
+- If lock is held by another plan, reconciler returns and requeues (standard controller-runtime pattern)
+- First to acquire lock wins; execution order is **not guaranteed**
```
jakefhyde: We will have to do something for etcd restore here. We could yield when a plan is failed, but things like k8s upgrades can "fail" and then spontaneously succeed, in the case of say a user messing around with security groups. In this case, you want to be able to stop reconciling an upgrade plan, run an etcd restore, and then go back to your reconciliation of the whole cluster.

docs/proposals/20250115-plan-improvements.md
```diff
@@ -0,0 +1,328 @@
...
+
+To address the 1MiB Secret/CRD size limit when plans contain large files (e.g., container images), the CRD supports content references:
+
+```yaml
+spec:
+  plan:
+    files:
+      - path: "/var/lib/rancher/images/core.tar"
+        contentRef:
+          oci: "registry.example.com/files/core:v1"
```
jakefhyde: I think it would be a good idea to expand on how we expect a user to instruct an orchestrator to leverage this (unless we aren't planning to do so, in which case I'm curious on how we are planning to leverage this).

PR #270, opened 2026-03-26
align system-agent image publishing for signed releases

jferrazbr:
## Problem

  The system-agent release flow was not aligned with the current signed image publishing pattern used in related repos. 

  ## Solution

  This change aligns system-agent image publishing with the current shared release pattern while preserving the existing image tags and manifest layout.

  The repo-owned Makefile targets were refactored so the main and SUC images each have distinct standard and signed-release publish paths. The release workflow was updated to use those paths for both the arch-specific tags and the merged multi-arch tags, without changing the image names or tag formats consumed downstream.

  ## Testing

  Validated the updated publishing flow locally against a throwaway registry to confirm:
  - existing arch-specific tags are preserved
  - merged tags are still created as multi-arch manifests
  - the signed-release path produces the expected attached metadata

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -16,7 +16,7 @@ env:
 jobs:
   # Build Docker images using multi-stage Dockerfile
   release:
-    name: Build and push images
+    name: Build and push arch-specific images
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


macedogm (dismissed):
From a high level view, the PR LGTM, but let's also have more +1 and a review from @tashima42.

tashima42 (changes requested):

tashima42 (approved):

macedogm (approved):

tashima42 (approved):

thatmidwesterncoder (approved):

jakefhyde (approved):

PR #274, opened 2026-04-09
setup for immutable releases

thatmidwesterncoder:


jakefhyde (dismissed):

.github/workflows/release.yaml
```diff
@@ -252,11 +259,18 @@ jobs:
...
         env:
-          GH_TOKEN: ${{ steps.app-token.outputs.token }}
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
-          gh release create ${{ github.ref_name }} \
-            --verify-tag \
-            --generate-notes \
-            dist/*
+          flags="--draft=false"
+          if [[ "${{ env.TAG }}" == *"-rc"* ]]; then
```
snasovich: Let's update it to cover other pre-release tags in case we start doing alphas or something.
Ideally want to do semver parsing but for now simpler check like "contains `-`" would do. It false-positives on things like `v0.1.1+build-1` but chances of us doing metadata tags (with `-` to boot) are low.

snasovich (dismissed):

.github/workflows/release.yaml
```diff
@@ -252,11 +259,18 @@ jobs:
           VERSION: ${{ github.ref_name }}
           COMMIT: ${{ github.sha }}
 
-      - name: Create GitHub release
+      - name: Add artfiacts to the GitHub release
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          gh release upload ${{ env.TAG }} dist/*
```
jferrazbr: do you think of adding this [--clobber option](https://cli.github.com/manual/gh_release_upload) is worth it? In case `gh release edit` fails in the next step and we need to run the job again

thatmidwesterncoder: Updated to pass that option as well.

jferrazbr (dismissed):

.github/workflows/release.yaml
thatmidwesterncoder: Big agree - spent some time trying to come up with a "clean" semver grep but that is probably more brittle and harder to update. Updated the if statement to just check against `*"-"*` now which will catch most things.

jferrazbr (approved):

HarrisonWAffel (approved):

jakefhyde (approved):

PR #275, opened 2026-04-10
Add integration test infrastructure and initial system-agent tests

furkatgofurov7:
Adds integration tests that exercise the system-agent's k8splan watcher against a real Rancher instance using the existing v2prov provisioning-tests framework. Tests run on a GHA native runner (no Dapper) against `rancher/rancher:v2.14-head`.

**CI infra setup + scripts:**
- New GHA workflow that sets up k3s binaries, Docker with cgroupfs + insecure-registry, and runs tests via `make integration-tests`.
- `fetch-provisioning-tests` script - clones rancher/rancher, extracts the rancher binary from the container image, and tags it as `rancher/rancher:dev` so provisioning-tests skips building from source.
- `integration-tests` script - builds system-agent from the current branch, copies integration tests into rancher's v2prov test tree (stripping `//go:build ignore` tags), patches provisioning-tests to use local binaries

**Design decisions**
- Tests live in system-agent with `//go:build ignore` tags so they don't break go build here. They import `github.com/rancher/rancher/tests/v2prov` packages from rancher's root module, which would pull in the entire rancher dependency tree if added to system-agent's go.mod file. The CI script strips the build tag after copying them into Rancher's module tree.
- Tests run alongside existing v2prov **MP/prebootstrap** tests under the same Rancher instance to avoid spinning up a second cluster, keeping CI under the 2-hour timeout.
- Rancher binary is pulled from a container image rather than built from source - avoids compiling all of rancher in CI and keeps the focus on testing system-agent changes.
- Timeouts are increased from upstream defaults (300s -> 600-1200s) because GHA ubuntu-latest runners have only 2 cores, which is a limitation that most probably can be solved by reaching out to the infra team, but for the initial phase, it should be sufficient.

Two system-agent-specific integration tests added:
1. provisions a single-node cluster and validates that the system-agent correctly populates plan secret status keys 
2. provisions a cluster, restarts the system-agent via systemctl, and verifies last-apply-time changes and the cluster stays Ready.

Docs:
- A new README was added under test/integration folder which explain test architecture, answering some "why"s' and "how"s' to add new integration tests in the future.

Note: I am planning to add more integration tests in follow-up PRs if the current approach seems fine and accepted by reviewers.

Part of: #260

### DIFF
--- a/.github/workflows/integration-tests.yaml
+++ b/.github/workflows/integration-tests.yaml
@@ -1,0 +1,86 @@
+name: Integration Tests
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, labeled, unlabeled]
+  workflow_dispatch:
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+jobs:
+  integration:
+    name: Integration Tests
+    runs-on: ubuntu-latest
+    timeout-minutes: 120
+    permissions:
+      contents: read
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+        with:
+          fetch-depth: 0
+      - name: Set up Go
+        uses: actions/setup-go@40f1582b2485089dde7abd97c1529aa768e1baff # v5
+        with:
+          go-version-file: 'go.mod'
+          check-latest: true
+          cache: true
+      - name: Configure Docker for cgroupfs/insecure-registry
+        run: |
+          echo '{"exec-opts": ["native.cgroupdriver=cgroupfs"], "insecure-registries": ["172.17.0.2:5000"]}' | sudo tee /etc/docker/daemon.json
+          if [ -d /run/systemd/system ]; then
+            sudo systemctl restart docker
+          fi
+          sudo docker info
+          # Alias 172.17.0.2 onto docker0 so the registry-cache is reachable.
+          sudo ip addr add 172.17.0.2/32 dev docker0
+      - name: Set up k3s binaries and CNI plugins
+        env:
+          CATTLE_K3S_VERSION: v1.35.3-k3s1
+        run: |
+          container_id=$(sudo docker create rancher/k3s:${CATTLE_K3S_VERSION})
+          for bin in blkid bandwidth cni conntrack containerd containerd-shim-runc-v2 ethtool firewall ip ipset k3s losetup pigz runc; do
+            sudo docker cp "${container_id}:/bin/${bin}" /usr/bin/ || true
+          done
+          sudo docker cp "${container_id}:/bin/aux/xtables-legacy-multi" /usr/bin/ || true
+          sudo docker rm "${container_id}"
+          sudo ln -sf /usr/bin/cni /usr/bin/bridge
+          sudo ln -sf /usr/bin/cni /usr/bin/flannel
+          sudo ln -sf /usr/bin/cni /usr/bin/host-local
+          sudo ln -sf /usr/bin/cni /usr/bin/loopback
+          sudo ln -sf /usr/bin/cni /usr/bin/portmap
+          sudo ln -sf /usr/bin/k3s /usr/bin/crictl
+          sudo ln -sf /usr/bin/k3s /usr/bin/ctr
+          sudo ln -sf /usr/bin/k3s /usr/bin/k3s-agent
+          sudo ln -sf /usr/bin/k3s /usr/bin/k3s-etcd-snapshot
+          sudo ln -sf /usr/bin/k3s /usr/bin/k3s-server
+          sudo ln -sf /usr/bin/k3s /usr/bin/kubectl
+          sudo ln -sf /usr/bin/pigz /usr/bin/unpigz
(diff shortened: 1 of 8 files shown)


furkatgofurov7:
@jakefhyde @HarrisonWAffel this should be ready for initial review when you get a time, thanks.

.github/workflows/integration-tests.yaml
```diff
@@ -0,0 +1,86 @@
...
+          sudo ln -sf /usr/bin/xtables-legacy-multi /usr/bin/ip6tables-save
+          sudo ln -sf /usr/bin/xtables-legacy-multi /usr/bin/ip6tables-restore
+      - name: Pre-pull container images
+        env:
+          CATTLE_AGENT_IMAGE: rancher/rancher-agent:v2.14-head
+          RANCHER_IMAGE: rancher/rancher:v2.14-head
+        run: |
+          sudo docker pull "$RANCHER_IMAGE" &
+          sudo docker pull "$CATTLE_AGENT_IMAGE" &
+          wait
```
anmazzotti, line 71: nitpick: is the wait needed and can these be just 2 commands instead?

scripts/integration-tests
```diff
@@ -0,0 +1,107 @@
...
+    while true; do
+        for pod in $(kubectl --kubeconfig "$KUBECONFIG" get pods -A -l rke.cattle.io/machine=true \
+            -o jsonpath='{range .items[*]}{.metadata.namespace}/{.metadata.name}{"\n"}{end}' 2>/dev/null); do
+            ns="${pod%%/*}"
+            name="${pod##*/}"
+            marker="/tmp/.sa-stream-${ns}-${name}"
+            if [ ! -f "$marker" ]; then
+                touch "$marker"
+                kubectl --kubeconfig "$KUBECONFIG" -n "$ns" exec "$name" -- \
+                    journalctl -u rancher-system-agent -f --no-pager 2>/dev/null \
```
anmazzotti, line 104: I miss where this stream should appear, I couldn't find it in the runner logs.
I was wondering if they can be saved in _artifacts instead.

Same applies to the `get events` above.

scripts/integration-tests
```diff
@@ -0,0 +1,107 @@
...
+
+cd "$(dirname "$0")/.." || exit 1
+
+SYSTEM_AGENT_DIR="$(pwd)"
+
+if [ "$V2PROV_TEST_DIST" != "rke2" ] && [ "$V2PROV_TEST_DIST" != "k3s" ]; then
+    export V2PROV_TEST_DIST=rke2
+fi
+
+if [ -z "${V2PROV_TEST_RUN_REGEX}" ]; then
```
anmazzotti, line 15: nitpick: a little comment explicitly mentioning that ProvisioningMP and PreBootstrap tests are ran from Rancher (with swapped agent binary), while SystemAgent are from here, would help. Also reinforcing the `Test_SystemAgent_*` naming convention for the future.

anmazzotti (dismissed):
Amazing work, also thank you for all the trail of comments, it helps a lot.

scripts/integration-tests
```diff
@@ -0,0 +1,107 @@
...
+set -ex
+
+make build
+
+mkdir -p /usr/share/rancher/ui/assets
+cp "${SYSTEM_AGENT_DIR}/install.sh" /usr/share/rancher/ui/assets/system-agent-install.sh
+
+source ./scripts/fetch-provisioning-tests
+
+# Copy system-agent integration tests into rancher's v2prov test tree.
```
anmazzotti, line 30: Another comment explicitly mentioning the `//go:build ignore` workaround and its need would help here.
This is probably a common mistake that will happen when adding new tests.

jakefhyde (dismissed):

.github/workflows/integration-tests.yaml
furkatgofurov7: @anmazzotti yes, `wait` is needed because those `docker pull` commands are being backgrounded with `&` and pulls are done in parallel

scripts/integration-tests
furkatgofurov7: Thanks, done

scripts/integration-tests
furkatgofurov7: Thanks, done. I also added a small documentation on how integration tests structure works and how to add a new ones (what to do and pay attention to) in https://github.com/rancher/system-agent/pull/275/commits/0a9a4dff7ad5a5481d3be01db51646b48db7c39d commit, thanks for suggestions

scripts/integration-tests
furkatgofurov7: Should be fixed now

anmazzotti: @furkatgofurov7 I still only see `rancher.log` in the latest collected artifacts.
Not really sure why, the code checks out for me. At very least I would expect the `events.log` to be present, since that's a very simple selection and it should always contain data.

anmazzotti: Unresolving the comment for visibility, but not a blocker for the PR.

anmazzotti (approved):

salasberryfin (approved):

jakefhyde (approved):

scripts/integration-tests
furkatgofurov7: @anmazzotti I will follow up on this in a separate PR, thanks!

PR #281, opened 2026-04-28
Consume plan API from rancher

### LINKED ISSUE #279: Utilize new pkg/plan from r/r
A new package is available, the existing logic in system-agent must be adapted to use it. This will include removing the types, implement inline that are now replaced by the ones exposed by rancher/rancher/pkg/plan 

The logic that watches and applies the plans defined by these structs remains unchanged in system-agent but is adapted to understand the new types.

Note: a new package is not out with r/r release since it was merged into main, an meantime we can workaround it by using r/r commit sha when consuming it from system agent and once it is part of rancher 2.15.0-alpha release replaced with actual release and PR might need to be hold until then.

furkatgofurov7:
Fixes: #279
Part of: #265


anmazzotti (dismissed):
I do believe that this package should live here.
Still welcome to be separate so that users don't need to import many indirect dependencies.

What worries me is that by moving it to `rancher/rancher`, any API change needs to be first implemented there and published, before it can be used by the system-agent, which is not an optimal situation. I also believe that architecturally the Plan API belongs to the system-agent, not Rancher.

However, since it was decided in the RFD to move it there, I'll just approve.

jakefhyde (dismissed):
This is extremely nitpicky, but WDYT about using `planapi` or `apiplan` instead of `planpkg`?

furkatgofurov7:
> This is extremely nitpicky, but WDYT about using planapi or apiplan instead of planpkg?

sure, replaced with `planapi` here and also in https://github.com/rancher/rancher/pull/54851 👍🏼

anmazzotti (approved):

jakefhyde (approved):

salasberryfin (approved):

PR #285, opened 2026-05-11
export applyinator byte functions

salasberryfin:
# Description

The e2e test framework added in rancher/system-agent#259 introduced `test/framework/decode.go`, which re-implements that same gzip+base64+JSON encoding/decoding logic directly in test code (`GetOutputMap`, `DecodePeriodicOutput`). As noted [here](https://github.com/rancher/system-agent/pull/259#discussion_r2883677551), this means the tests are using a parallel implementation of the encoding/decoding logic. If the production encoding logic ever changes, the tests could silently diverge and continue to pass while verifying incorrect behavior.

By exporting the existing functions, they can be re-used and properly tested.


jakefhyde (dismissed):

salasberryfin:
Closing this as it creates a dependency on the `applyinator` module on `test/framework/decode.go` that triggers an error on integration tests.

Considering this is simply an improvement on testability and maintainability, it is not worth the effort required to re-think how integration tests are executed, which is already complex enough.

PR #313, opened 2026-06-17
[2.13] bump the fallback version in install.sh to v0.3.16

thatmidwesterncoder:


install.sh
```diff
@@ -46,7 +46,10 @@ fi
 #   - CATTLE_AGENT_STRICT_VERIFY | STRICT_VERIFY (default: false)
 #   - CATTLE_AGENT_FALLBACK_PATH (default: )
 
-FALLBACK=v0.3.13
+FALLBACK=v0.3.16
```
jiaqiluo: Can you bump it to the latest v0.13.x tag?

jiaqiluo (changes requested):

PR #332, opened 2026-06-23
Ignore the deprecated 'release/v0.3' branch in Renovate

jiaqiluo:
There is no need to generate PRs for bumping dependencies in the deprecated 'release/v0.3' branch.
This PR updates the branch pattern to match any string that starts with the word `release/`, except if that version is exactly v0.3.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -3,8 +3,8 @@
     "github>rancher/renovate-config#release"
   ],
   "baseBranchPatterns": [
-      "$default",
-      "/^release\\/.*/"
+    "$default",
+    "/^release\\/(?!v0\\.3$).*/"
   ],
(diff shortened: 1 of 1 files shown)


salasberryfin (commented):
Should we ignore old branches or [explicitly set the ones we would like to cover](https://github.com/rancher/system-agent/pull/341)? In the end both options require periodic manual changes as existing release branches will at some point be deprecated.

thatmidwesterncoder:
honestly i would be very much in favor of just deleting the branch since we didn't seem to use it (hasn't been updated since ~2023).

jiaqiluo:
Close in favor of https://github.com/rancher/system-agent/pull/341

PR #341, opened 2026-06-25
set explicit renovate branches

salasberryfin:
# Description

Renovate is creating PRs for all release branches, including old ones like `release/v0.3`. This PR explicitly configures the target branches.

We could have a dynamic configuration with a more complex regular expression but I think we can just add one when a new branch is created, as it doesn't happen that often (`rancher/rancher` seems to do the same).

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -4,7 +4,8 @@
   ],
   "baseBranchPatterns": [
       "$default",
-      "/^release\\/.*/"
+      "release/v2.13",
+      "release/v2.14"
   ],
   "prHourlyLimit": 2
 }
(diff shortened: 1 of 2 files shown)


jiaqiluo (changes requested):
Can you also update the readme to mention that we will need to update this list when a new release branch is added or an old one is deprecated?

salasberryfin:
Thanks @jiaqiluo. I added a note to the readme.

jiaqiluo (approved):

HarrisonWAffel (approved):

.github/renovate.json
```diff
@@ -4,7 +4,8 @@
   ],
   "baseBranchPatterns": [
       "$default",
-      "/^release\\/.*/"
+      "release/v2.13",
+      "release/v2.14"
```
jiaqiluo, line 8: I think we will soon have a "release/v2.15" branch.

jiaqiluo (changes requested):
Hi @salasberryfin, can you add  "release/v2.15" branch to the list?

PR #393, opened 2026-07-13
fix: Skip proxy vars in SUC env resurrection so deletion propagates

pratikjagrut:
## Problem

Deleting an agent env var (`HTTP_PROXY`/`HTTPS_PROXY`/`NO_PROXY`) on a downstream cluster doesn't propagate — the old value  persists in `/etc/systemd/system/rancher-system-agent.env`.

Fixes SURE-6828 / rancher/rancher#45210

## Cause

`package/suc/run.sh` re-exports vars from the existing env file before running `install.sh`. On delete, the plan correctly omits the var, but the loop resurrects the stale value from the old file and `create_env_file` writes it back.

## Fix

Skip proxy vars in the loop so the SUC plan env is the source of truth (present → written, absent → cleared). `PATH` and
other vars still preserved. Also fix `${line##*=}` → `${line#*=}` for values containing `=`.

## Testing

Set → update → delete `NO_PROXY` on a downstream cluster with the patched `-suc` image: value now clears on delete. No regression for set/update/no-proxy cases.

### DIFF
--- a/package/suc/run.sh
+++ b/package/suc/run.sh
@@ -45,14 +45,26 @@ export CATTLE_AGENT_UNINSTALL_LOCAL=true
 export CATTLE_AGENT_BINARY_LOCAL_LOCATION=${TMPDIR}/rancher-system-agent
 export CATTLE_AGENT_UNINSTALL_LOCAL_LOCATION=${TMPDIR}/rancher-system-agent-uninstall.sh
 if [ -s /host/etc/systemd/system/rancher-system-agent.env ]; then
-  for line in $(grep -v '^#' /host/etc/systemd/system/rancher-system-agent.env); do
+  while IFS= read -r line; do
+    case "$line" in
+      '#'*|'') continue ;;
+    esac
     var=${line%%=*}
-    val=${line##*=}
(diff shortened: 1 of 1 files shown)


package/suc/run.sh
```diff
@@ -45,14 +45,26 @@ export CATTLE_AGENT_UNINSTALL_LOCAL=true
 export CATTLE_AGENT_BINARY_LOCAL_LOCATION=${TMPDIR}/rancher-system-agent
 export CATTLE_AGENT_UNINSTALL_LOCAL_LOCATION=${TMPDIR}/rancher-system-agent-uninstall.sh
 if [ -s /host/etc/systemd/system/rancher-system-agent.env ]; then
```
pedromfcarvalho: It will be simpler to remove this whole if block entirely.

The only variables that `install.sh` writes to the env file are the proxy variables anyways, and potentially the adjusted `PATH` like you mentioned, but this is unlikely to be ever empty here so it won't be exported, and it will be re-adjusted by `install.sh` anyways.

pedromfcarvalho (approved):

PR #407, opened 2026-08-07
[Main] code refactor and bug fixes

jiaqiluo:
Issue: 
- https://github.com/rancher/rancher/issues/56594

Only for the main branch. 
This PR refactors the codebase to make it more modular, testable, and easier to maintain long-term. 

It also fixes a few bugs identified during the refactoring process. Please see the separate reviewer comments for more details.

The [Copilot summary](https://github.com/rancher/system-agent/pull/407#pullrequestreview-4885096121) is really helpful for understanding the changes. 

All tests, including the integration tests, passed 🥳


pkg/localplan/watcher.go
```diff
@@ -24,6 +24,16 @@ func WatchFiles(ctx context.Context, applyinator applyinator.Applyinator, bases
...
 
+	// Nothing else creates the local plan directory -- install.sh writes localPlanDirectory into
+	// the config but only mkdirs the interlock, config, and bin dirs. Create it here so a missing
+	// directory does not make listFiles report a walk error on every 5s poll forever. 0700 matches
+	// appliedPlanDir: plans can carry secrets.
+	for _, base := range bases {
+		if err := os.MkdirAll(base, 0700); err != nil {
+			logrus.Errorf("[localplan] Failed to create local plan directory %s: %v", base, err)
+		}
+	}
```
jiaqiluo, line 37: **To Reviewer:** 

WatchFiles now creates each base with MkdirAll before starting the poll loop, which makes main.go's existing warning accurate. 0700 matches appliedPlanDir in applyinator; plans can carry secrets. MkdirAll failures are logged rather than returned, since WatchFiles has no error return and listFiles will surface the condition anyway.

pkg/applyinator/applyinator.go
```diff
@@ -337,19 +359,69 @@ func (a *Applyinator) Apply(ctx context.Context, input ApplyInput) (ApplyOutput,
...
+	nowUnixTimeString := now.Format(time.UnixDate)
+	restartPendingInterlockFilePath := filepath.Join(a.interlockDir, restartPendingInterlockFile)
+	applyinatorActiveInterlockFilePath := filepath.Join(a.interlockDir, applyinatorActiveInterlockFile)
+
+	// First off, remove check and remove the active interlock as the applyinator is not actually active
+	if _, err := os.Stat(applyinatorActiveInterlockFilePath); err == nil {
+		if err := os.Remove(applyinatorActiveInterlockFilePath); err != nil {
+			logrus.Errorf("[applyinator] unable to remove applyinator active interlock file %s: %v", applyinatorActiveInterlockFilePath, err)
+		}
+	}
```
jiaqiluo, line 397: **To Reviewer:** 
now it removes the  file at `applyinatorActiveInterlockFilePath` (value: "{Interloc-dir}/applyinator-active", instead of the bare relative filename `applyinatorActiveInterlockFile`( value: "applyinator-active")

pkg/applyinator/applyinator.go
```diff
@@ -484,40 +561,46 @@ func (a *Applyinator) execute(ctx context.Context, prefix, executionDir string,
...
+		stderrBuffer bytes.Buffer
 	)
 
+	stdoutTarget := &stdoutBuffer
+	stderrTarget := &stderrBuffer
+	stdoutLock := &sync.Mutex{}
+	stderrLock := stdoutLock
+
 	if combinedOutput {
-		stderrBuffer = stdoutBuffer
```
jiaqiluo, line 508: **Note to reviewers:**

The bug in the original code: 

`stderrBuffer = stdoutBuffer` copied an empty bytes.Buffer by value, so combinedOutput=true (used for one-time instructions) never actually merged stdout/stderr; stderr was silently dropped from saved one-time output.

pkg/applyinator/applyinator.go
```diff
@@ -527,15 +610,19 @@ func (a *Applyinator) execute(ctx context.Context, prefix, executionDir string,
 	// Wait for I/O to complete before calling cmd.Wait() because cmd.Wait() will close the I/O pipes.
 	_ = eg.Wait()
 	exitCode := 0
-	if err := cmd.Wait(); err != nil {
-		if ee, ok := err.(*exec.ExitError); ok {
```
jiaqiluo, line 531: **Note to reviewers:**

The bug in the original code:
`cmd.Wait()`'s error was shadowed within its own if statement and never reached the return value.

.github/workflows/integration-tests.yaml
```diff
@@ -63,8 +63,8 @@ jobs:
           sudo ln -sf /usr/bin/xtables-legacy-multi /usr/bin/ip6tables-restore
       - name: Pre-pull container images
         env:
-          CATTLE_AGENT_IMAGE: rancher/rancher-agent:v2.15-0b6b24d9811ed8dbe02ab315847602f7facf2244-head
-          RANCHER_IMAGE: rancher/rancher:v2.15-0b6b24d9811ed8dbe02ab315847602f7facf2244-head
+          CATTLE_AGENT_IMAGE: rancher/rancher-agent:v2.15.0
```
jakefhyde: Just curious, any idea why we were at this tag to begin with? Seems like a good chore task might be to automate bumping to 2.x.0 whenever rancher release a new minor.

test/framework/const.go
```diff
@@ -32,8 +32,9 @@ const (
 	// ShortTestLabel is used to filter tests for PR gate (fast tests).
 	ShortTestLabel = "short"
 
-	// FullTestLabel is used for nightly (all tests).
-	FullTestLabel = "full"
```
jakefhyde, line 36: Just curious, any reason why this was changed? I know it was likely based on https://github.com/rancher/turtles/blob/main/test/e2e/const.go#L197, but I don't think it's necessarily a convention.

pkg/prober/status_test.go
```diff
@@ -0,0 +1,119 @@
...
+			succeeded:        false,
+			successThreshold: 1,
+			failureThreshold: 3,
+			want:             planapi.ProbeStatus{Healthy: false, SuccessCount: 0, FailureCount: 1},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
```
jakefhyde, line 111: IIRC you have to do the following:
```suggestion
	for _, tt := range tests {
	  tt := tt
		t.Run(tt.name, func(t *testing.T) {
			t.Parallel()
```

jakefhyde, line 111: IIRC you have to do the following:
```suggestion
	for _, tt := range tests {
	  tt := tt
		t.Run(tt.name, func(t *testing.T) {
			t.Parallel()
```

jakefhyde, line 111: IIRC you have to do the following:
```suggestion
	for _, tt := range tests {
	  tt := tt
		t.Run(tt.name, func(t *testing.T) {
			t.Parallel()
```

test/README.md
```diff
@@ -2,10 +2,44 @@
...
+│   ├── const.go               # namespace/label/timeout constants, ShortTestLabel/LongTestLabel
+│   ├── cluster.go             # kubectl wrappers (apply, wait-for-ready, exec, logs, HTTP test server lifecycle)
+│   ├── plan.go                # PlanBuilder fluent API for constructing plan JSON
+│   ├── secret.go              # plan Secret CRUD + polling/Eventually helpers
+│   ├── decode.go               # gzip+base64+JSON decoding of applied-output/periodic-output fields
+│   └── template.go            # `${VAR}` template rendering for the embedded manifests
+├── testenv/                    # Kind cluster + agent lifecycle (used only by suite_test.go)
+│   ├── setup.go                # SetupTestCluster (Kind + image load + RBAC), DeployRemoteAgent (SA token, kubeconfig, DaemonSet)
+│   └── cleanup.go              # CleanupTestCluster (tears down the Kind cluster)
+└── integration/                 # separate, non-Ginkgo suite — see integration/README.md
```
jakefhyde: are these indented correctly?

pkg/k8splan/secret_outcome.go
```diff
@@ -0,0 +1,67 @@
...
+// buildSecretDataUpdates computes the Secret data key writes resulting from one apply, mirroring
+// the pre-refactor post-apply mutation block in the OnChange closure. The caller is responsible
+// for merging the returned map into secret.Data and emitting the returned logs.
+func buildSecretDataUpdates(in applyOutcomeInput) (map[string][]byte, []decisionLog) {
+	updates := map[string][]byte{
+		AppliedPeriodicOutputKey: in.PeriodicOutput,
+	}
+
+	failed := (in.NeedsApplied && !in.OneTimeApplySucceeded) || (!in.NeedsApplied && in.WasFailedPlan)
+	if failed {
```
jakefhyde, line 32: This is a general comment, but I noticed this when working on the r/r side: I feel like we have too many error cases that have to be distinctly parsed so we can actually derive what's going on. I think this lays some good groundwork, but I'd like to build out the API in the future so we can no whether a plan is failed without having to check multiple cases to figure out why.

pkg/applyinator/applyinator.go
```diff
@@ -87,237 +88,258 @@ type ApplyInput struct {
...
+
+// periodicInstructionDue decides whether a periodic instruction should run now, given the
+// previously recorded output for that instruction. An unset or unparsable last-run timestamp is
+// treated as "no history" (always due). forced (the one-time instructions ran this cycle) bypasses
+// both the period and the failure cooldown.
+func periodicInstructionDue(now time.Time, prev planapi.PeriodicInstructionOutput, periodSeconds int, forced bool) (due bool, failures int) {
+	if t, ok := parseUnixTimeOrZero("last successful run time", prev.LastSuccessfulRunTime); ok {
+		effectivePeriod := periodSeconds
+		if effectivePeriod == 0 {
+			effectivePeriod = 600
```
jakefhyde: Can you make this a const

pkg/applyinator/applyinator.go
```diff
@@ -87,237 +88,258 @@ type ApplyInput struct {
...
+			logrus.Debugf("[applyinator] not running periodic instruction as period duration has not elapsed since last successful run")
+			return false, failures
+		}
+	}
+
+	if prev.LastFailedRunTime != "" {
+		if t, ok := parseUnixTimeOrZero("last failed run time", prev.LastFailedRunTime); ok {
+			failures = prev.Failures
+			failureCooldown := failures
+			if failureCooldown > 6 {
```
jakefhyde: can you make this a const as well

jakefhyde (changes requested):

pkg/prober/status_test.go
jakefhyde: Resolving, apparently this has been fixed since go 1.22 :O

.github/workflows/integration-tests.yaml
jiaqiluo: The commit `0b6b24d98` from the original tag is where the `plan-state` was implemented on the Rancher side:
https://github.com/rancher/rancher/commit/0b6b24d9811ed8dbe02ab315847602f7facf2244

Yes, you’re right. This is a chore that needs to be done regularly, especially when new provisioning test cases are added to Rancher and we want to run them in System Agent as well.

We should automate this process, and I’ll create an issue to track it.

test/framework/const.go
jiaqiluo: Good to know where the term comes from.

Neither the old `full` tag nor the new `long` tag is currently used anywhere; all existing E2E test cases are labeled as `short`. So, `LongTestLabel` is more of an indicator for potential future use.

The rename from `full` to `long` was mainly to make the terminology sound more natural as the opposite of `short`.

Let me know if it is better to switch back to "full".

pkg/prober/status_test.go
jiaqiluo: Before Go 1.22, loop vars in for range were reused across iterations. Closures (like subtests) could all observe the same final value unless you did one of these:
- tt := tt inside the loop, or
- passed tt as a parameter to the closure.

That’s why this pattern became idiomatic in tests, especially with t.Parallel(): parallel subtests start later, so by then the loop may have advanced, exposing the capture bug.

From Go 1.22+, range variables are per-iteration by default, so the extra rebinding is usually redundant.

https://go.dev/blog/loopvar-preview

test/README.md
jiaqiluo: Yes, I can confrim that the identification is correct in the local editor.

pkg/applyinator/applyinator.go
jiaqiluo: Fixed.

pkg/applyinator/applyinator.go
jiaqiluo: Fixed.

pkg/prober/status_test.go
jiaqiluo: See https://github.com/rancher/system-agent/pull/407#discussion_r3777762282

.github/workflows/integration-tests.yaml
jiaqiluo: Here is the issue:
- https://github.com/rancher/rancher/issues/56621

jakefhyde (dismissed):

pkg/config/config_windows.go
```diff
@@ -7,12 +7,12 @@ import (
 	"os"
 )
 
-// pathOwnedByCurrentUser is abstracted out with Windows being a no op.
+// pathOwnedByCurrentUser validates ownership on Unix. On Windows return nil.
```
HarrisonWAffel: Very small non-blocking nit: This is `config_windows.go`, it should describe the behavior on windows exclusively, I think the old comment was fine

test/README.md
HarrisonWAffel: The tree itself is indented correctly, but the `#` comments are a bit messy when viewed in githubs rich preview. Nothing to block this PR on though

pkg/config/config_test.go
```diff
@@ -0,0 +1,206 @@
...
+func TestPathOwnedByCurrentUser(t *testing.T) {
+	t.Parallel()
+
+	path := writeConfigFile(t, "owner-test", []byte("{}"), 0600)
+	if err := pathOwnedByCurrentUser(path); err != nil {
+		t.Errorf("expected a file created by this process to be owned by the current user, got: %v", err)
+	}
+}
+
+// TestParseRejectsFileOwnedByAnotherUser covers the ownership branch of Parse. CLAUDE.md calls out
```
HarrisonWAffel: It doesn't look like this PR adds a standard `CLAUDE.md`, so this comment should be trimmed IMO

```
// TestParseRejectsFileOwnedByAnotherUser covers the ownership branch of Parse.
// Chowning to another user requires root, so this is skipped otherwise.
```

pkg/k8splan/decision_log.go
```diff
@@ -0,0 +1,58 @@
...
+	decisionDebug
+	decisionInfo
+	decisionError
+)
+
+// decisionLog is one log line produced while evaluating a plan decision.
+// Decision functions are pure for unit testing.
+// The collected log lines are the primary field-debugging surface on provisioned nodes.
+// Tests can assert on the collected messages.
+type decisionLog struct {
```
HarrisonWAffel: I kind of feel like we're reinventing the logging wheel here, can't we just use [a child logger](https://www.dash0.com/guides/logrus-golang#child-loggers) to prefix these fields? 

I'm generally against unit testing error logs unless they're apart of an established contract with another components / package, or contain runtime information that needs to be tested. But I think this pattern could be a footgun too. We're now requiring that future maintainers remember to call `emitDecisionLogs` in the right places, and if that gets missed we'll completely drop relevant logs. We also now need to pass a slice of logs across functions.

In this case I think we'll just be making more work for ourselves when making slight logging changes in the future, and opening the potential for logs to be dropped all together. 

Let me know if I'm missing something about the benefits of this approach though

HarrisonWAffel (changes requested):
My only real comment is on how we're handling logs, and a few other nits. In general this looks like a great change and matches the existing behavior AFAICT

test/README.md
jiaqiluo: Fixed by moving the # comments to their own section under the tree.

pkg/config/config_windows.go
jiaqiluo: Fixed by reverting the change

pkg/config/config_test.go
jiaqiluo: Fixed.

pkg/k8splan/decision_log.go
jiaqiluo: Thank you for the feedback. It makes sense to me.

The code was simplified significantly in the latest commit: the `decisionLog` was removed, and the unit tests no longer validate log strings, since they are not part of any contract.

It should be more developer-friendly now. :) 

Please check the latest two commits for the changes.

HarrisonWAffel (approved):

jiaqiluo:
Hi @jakefhyde, could you please review the PR again? The changes are included in the latest two commits since your previous approval.

jakefhyde (approved):

Issue #403, opened 2026-08-04
[Enhancement] A system-agent killed mid-plan leaves interlock files and a stale-resourceVersion wedge that never self-heal


marytlf:
## Describe the bug

When the `rancher-system-agent` process is terminated in the middle of applying a plan — as happens during an RKE2 upgrade on a node whose etcd is briefly unresponsive — the agent leaves behind state that no later invocation ever cleans up:

1. The `applyinator-active` interlock file is leaked, because the code intended to clear a stale one checks the **wrong path** and the `defer` that would remove it never runs on SIGTERM.
2. The `restart-pending` interlock file is leaked, because `install.sh` writes it with a bare `touch` and only removes it on the happy path.
3. The plan secret reconcile wedges on a `secret received was too old` error, because a resourceVersion that moves backwards after an informer relist is treated as a permanent error rather than a transient stale read.

Individually each guard is defensible. Together they turn a transient storage stall into a state that requires **manual intervention** (deleting the plan) to escape, long after the underlying condition has cleared.

The triggering storage latency is not a system-agent bug and is out of scope here. What this issue proposes is that the agent should never require a human to unstick it from a state created by a killed predecessor.

## To Reproduce

Observed on a vSphere-backed downstream cluster provisioned via Terraform → Rancher → vSphere, upgrading RKE2 `v1.35.4+rke2r1` → `v1.35.6+rke2r1`, where the datastore stalled an etcd `fdatasync` for 3.4 s during the control-plane node's upgrade.

The failure does not require slow storage to reproduce. Each of the three defects can be triggered independently.

## Result

The observed timeline, with the log line each defect produces:

```
09:48:43   1st upgrade attempt begins, plan 0f064a47…
09:48:56   cattle-system_apply-system-agent-upgrader:
           "Active plan reconciliation detected. Sleeping for 5 seconds and retrying check"   <-- defect 1
07:48:59Z  etcd: apply request took too long, took 207.9ms, expected 100ms
07:49:00Z  etcd: leader failed to send out heartbeat on time … likely from slow disk
07:49:02Z  kube-apiserver: context deadline exceeded
07:49:03Z  etcd: "slow fdatasync", took 3.437s, expected 1s
09:49:07   plan 0f064a47… receives killed signal
09:49:07   warning: signal received: "terminated", canceling context...
09:49:09   2nd attempt, plan d8347aba…; plan 0f064a47… was never deleted
09:49:00   error: handler secret-watch: error encountered when running apply:
[...]

PR #431, opened 2026-08-31
Clear the active interlock on graceful shutdown

marytlf:
**Read PR part A and B to better understand the changes**: https://github.com/rancher/system-agent/pull/429 and https://github.com/rancher/system-agent/pull/430 
## Issue:  https://github.com/rancher/system-agent/issues/403

**PR A** and **PR B** make a leaked interlock *recoverable*. This PR stops it leaking in the first place on the ordinary shutdown path — the one that happens on every `systemctl restart`, and therefore the common case rather than the crash case.

## Problem

`Apply` removes `interlock/applyinator-active` in a `defer`, but nothing kept the process alive long enough for that `defer` to run. `main.go:127` returned as soon as `topContext` was cancelled:

```go
<-topContext.Done()
return nil
```

A `SIGTERM` arriving mid-apply therefore raced the cleanup, and on a `systemctl restart` the exit usually won. The next `install.sh` then found a file whose owner was already gone — the state PRs A and B exist to recover from.

## Solution

`Applyinator.Drain(ctx)` waits for any in-flight `Apply` — by taking `a.mu`, the same mutex `Apply` holds for its entire duration — and then guarantees the file is gone. `main.go` calls it after `<-topContext.Done()`.

Three properties worth review attention:

**It is bounded.** wrangler hard-exits on a second signal, so an open-ended wait would simply be killed. Hence the 30s cap.

**It runs on its own context.** `topContext` is already cancelled at the call site, so deriving from it would expire immediately and defeat the wait. `context.Background()` with an explicit timeout is deliberate.

**It clears the file on both paths, including timeout.** Giving up on the wait is precisely when the interlock is most likely to be stranded, so the removal sits after the `select`, not inside the success arm.

`Drain` is a no-op when `interlockDir` is empty, since interlocks are optional but shutdown is not.

This does **not** replace the self-heal in PRs A and B, which still covers `SIGKILL`, OOM and power loss. It removes the case that does not need to reach them.

### Alternatives considered

**Register the drain as a wrangler `OnShutdown` hook.** Would keep `main.go` untouched, but ordering against the controller factory's own shutdown is not guaranteed, and the drain must run after the watchers stop enqueuing work.

**Widen the `defer` in `Apply` to catch signals directly.** Signal handling belongs at the top level; `Apply` has no business installing handlers, and it would not help when the process exits from a different goroutine.

**Make the timeout configurable.** 30s is well inside systemd's default `TimeoutStopSec=90s`. A knob invites a value that exceeds it, at which point systemd `SIGKILL`s and the drain is worse than useless. Left as a constant.

### Known limitations

- Only covers *graceful* shutdown. `SIGKILL`, OOM and power loss still leak the file, by construction — that is what PRs A and B are for.
- If an apply legitimately runs longer than 30s (common: RKE2 upgrades), the drain times out and clears the interlock while the apply is still running. This is the right trade — the process is going away regardless, and a stranded file is the worse outcome — but it means the interlock is not a hard guarantee during shutdown.
- Adds up to 30s to `systemctl stop` when an apply is in flight. Well within `TimeoutStopSec`, but it is a visible change in shutdown time.

#### Manual Testing

Use the harness in `PR-A-owner-stamped-interlock.md`, with `git checkout fix-c-graceful-drain` (which includes A and B).

**C1 — The leak, before and after.** The whole point of the PR.

```sh
# On the parent branch first, to see the leak:
git checkout fix-b-restart-pending-owner && CGO_ENABLED=0 go build -o /usr/local/bin/rancher-system-agent .
rm -f /var/lib/rancher/agent/plans/test.pos
CATTLE_LOGLEVEL=debug rancher-system-agent sentinel &
sleep 10                                      # the 60s sleep instruction is running
ls -l /var/lib/rancher/agent/interlock/       # applyinator-active present
kill -TERM %1 && wait
ls -l /var/lib/rancher/agent/interlock/       # LEAKED

# On this branch:
git checkout fix-c-graceful-drain && CGO_ENABLED=0 go build -o /usr/local/bin/rancher-system-agent .
rm -f /var/lib/rancher/agent/plans/test.pos
CATTLE_LOGLEVEL=debug rancher-system-agent sentinel &
sleep 10
time kill -TERM %1 ; wait
ls -l /var/lib/rancher/agent/interlock/       # gone
```

Expect on this branch: shutdown pauses while the instruction finishes (up to 30s), `drain complete, no apply in flight` at debug level, and no file left behind.

**C2 — The timeout is real.** Make the plan outlast the cap:

```sh
cat > /var/lib/rancher/agent/plans/test.plan <<'EOF'
{"instructions":[{"name":"slow","command":"/bin/sh","args":["-c","sleep 300"],"saveOutput":true}]}
EOF
rm -f /var/lib/rancher/agent/plans/test.pos
CATTLE_LOGLEVEL=debug rancher-system-agent sentinel &
sleep 10
time kill -TERM %1 ; wait
ls -l /var/lib/rancher/agent/interlock/ 
```

Expect: exits after **~30s, not 300**, logs `drain timed out waiting for in-flight apply; clearing active interlock anyway`, and the file is gone regardless.

**C3 — Idle shutdown is not slowed.** `SIGTERM` with no plan in flight should return immediately. Worth timing, since a regression here would add 30s to every restart.

**C4 — `SIGKILL` still leaks, and is still recovered.** Confirms C is additive to A, not a replacement:

```sh
rm -f /var/lib/rancher/agent/plans/test.pos
CATTLE_LOGLEVEL=debug rancher-system-agent sentinel &
sleep 10 && kill -9 %1
ls -l /var/lib/rancher/agent/interlock/      # leaked, expected
rm -f /var/lib/rancher/agent/plans/test.pos
CATTLE_LOGLEVEL=debug rancher-system-agent sentinel   # PR A clears it
```

**C5 — Under systemd**, on a node with the agent installed:

```sh
rm -f /var/lib/rancher/agent/plans/test.pos
systemctl restart rancher-system-agent    # during an active plan apply
ls -l /var/lib/rancher/agent/interlock/
journalctl -u rancher-system-agent | grep -i drain
```

Check `systemctl show rancher-system-agent -p TimeoutStopSec` is comfortably above 30s (default 90s).

**C6 — No interlock configured.** Remove `interlockDirectory` from `config.yaml` and confirm shutdown does not panic.

#### Automated Testing

`pkg/applyinator/drain_test.go` — 5 tests: a leaked interlock is removed; a missing file does not panic; an empty `interlockDir` does not panic; **the drain actually waits** (asserts `Drain` has *not* returned 100ms into a held lock, then returns promptly once released); and the timeout path is bounded by its context deadline *and* still clears the file.

The waiting tests hold `a.mu` directly rather than running a real `Apply`, because executing an instruction chowns the working directory to root and unprivileged runners cannot.

Full suite green: `go build ./...`, `go vet ./...`, `gofmt -l`, `go test ./...`.

### QA Testing Considerations

- The interesting window is **`SIGTERM` arriving mid-apply** — an idle shutdown exercises none of this.
- Time `systemctl stop` and `systemctl restart` both idle and mid-plan; confirm neither approaches `TimeoutStopSec`.
- A long RKE2 upgrade plan will exceed 30s. Confirm the timeout path is clean: agent exits, interlock cleared, next start proceeds.
- Confirm the `SIGKILL` path still leaks *and* still recovers via PR A — C must be additive.

### Regressions Considerations

- **Shutdown latency.** Up to 30s added when an apply is in flight. If any packaging sets `TimeoutStopSec` below that, systemd will `SIGKILL` mid-drain — the interlock then leaks and PR A's self-heal handles it, so the failure is graceful, but the drain would be pointless there.
- **The goroutine taking `a.mu` is never released** if `ctx` expires first; it lives until the in-flight apply finishes, which is immediately before process exit. Harmless here, but it is a deliberate leak rather than an oversight.
- **Ordering in `main.go`.** `Drain` runs after all watchers have been signalled to stop. If a future change starts a watcher that can call `Apply` after `topContext` is cancelled, the drain could clear an interlock belonging to a *starting* apply.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -929,6 +929,24 @@ ensure_applyinator_not_active() {
     rm -f "${FILE_APPLYINATOR}"
 }
 
+write_interlock_owner() {
+    _boot=""
+    if [ -r /proc/sys/kernel/random/boot_id ]; then
+        _boot=$(cat /proc/sys/kernel/random/boot_id)
+    fi
+    # Format must match interlockOwner.marshal() in pkg/applyinator/interlock.go.
+    # LC_ALL=C keeps the day/month names parseable by Go's time.UnixDate, and the
+    # timestamp is written in UTC with a literal "UTC" zone: Go's UnixDate layout
+    # cannot parse a numeric zone (glibc %Z emits e.g. "+0530" where the zone has
+    # no abbreviation), and it resolves an unknown abbreviation like "CEST" to a
+    # zero offset, which would skew the timeout window.
+    # No start= field is written here: the agent treats a missing one as "unknown"
+    # and falls back to PID-only liveness, which is sufficient for the short,
+    # trap-guarded window the installer holds this file.
+    printf 'pid=%s\nboot=%s\ntime=%s\n' "$$" "${_boot}" "$(LC_ALL=C date -u '+%a %b %e %H:%M:%S UTC %Y')" > "$1"
+    chmod 0600 "$1"
+}
+
 do_install() {
     if [ $(id -u) != 0 ]; then
       fatal "This script must be run as root."
@@ -940,7 +958,11 @@ do_install() {
     ensure_directories
     verify_downloader curl || fatal "can not find curl for downloading files"
 
-    touch ${CATTLE_AGENT_VAR_DIR}/interlock/restart-pending
+    # Stamp the file with this shell's identity so the agent can tell a live
+    # installer from one that was killed, and remove it on any exit path — every
+    # fatal between here and the final rm previously leaked it.
+    write_interlock_owner ${CATTLE_AGENT_VAR_DIR}/interlock/restart-pending
+    trap 'rm -f ${CATTLE_AGENT_VAR_DIR}/interlock/restart-pending' EXIT INT TERM
     ensure_applyinator_not_active
 
     if [ -z "${CATTLE_CA_CHECKSUM}" ] && [ $(echo "${CATTLE_AGENT_STRICT_VERIFY}" | tr '[:upper:]' '[:lower:]') = "true" ]; then
--- a/install.sh
+++ b/install.sh
@@ -592,6 +592,7 @@ RestartSec=5s
 Environment=CATTLE_LOGLEVEL=${CATTLE_AGENT_LOGLEVEL}
 Environment=CATTLE_AGENT_CONFIG=${CATTLE_AGENT_CONFIG_DIR}/config.yaml
 Environment=CATTLE_AGENT_STRICT_VERIFY=${CATTLE_AGENT_STRICT_VERIFY}
+ExecStartPre=-/bin/rm -f ${CATTLE_AGENT_VAR_DIR}/interlock/applyinator-active
 ExecStart=${CATTLE_AGENT_BIN_PREFIX}/bin/rancher-system-agent sentinel
 EOF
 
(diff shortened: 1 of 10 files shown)


marytlf:
Closing since https://github.com/rancher/system-agent/pull/407 already is merged and take care of most of the issues.

PR #453, opened 2026-09-03
[2.16] plan pause and cancellation

jiaqiluo:
Issues:
- https://github.com/rancher/rancher/issues/54265
- https://github.com/rancher/rancher/issues/54255
 
## Summary

This PR adds support for **plan pause and cancellation** in `system-agent`'s remote mode (`pkg/k8splan`). 

Operators can now suspend or cancel executing plans directly via Kubernetes Secret annotations. This enables safe plan resumption from checkpoints and grants granular control over process termination behavior.

### Pause a Plan
Set `plan.cattle.io/paused = "true"` on the plan Secret.
* **Behavior:** The currently running instruction completes, but subsequent instructions are blocked. 
* **State:** A checkpoint is saved. 
* **Resume:** Remove the annotation or set it to `"false"`.

### Cancel a Plan
Set `plan.cattle.io/canceled = "true"` on the plan Secret.
* **Behavior:** Immediately terminates the in-flight instruction (`SIGTERM`/`SIGKILL` on Unix; Job Object termination on Windows). 
* **State:** Cancellation is **terminal**—canceled plans cannot be resumed directly. 
* **Restart:** The orchestrator must reset `plan-state` to `pending` or recreate the Plan secret.

### Resume a Paused Plan
Remove the `plan.cattle.io/paused` annotation or set its value to `"false"`.

### Annotation Handling
* **Valid values:** `"true"` and `"false"` (absent defaults to `"false"`).
* **Invalid values:** Handled as configuration errors. The plan halts, error status writes are skipped (preventing write-loop storms), and the workqueue retries using exponential backoff/rate limiting.
* **Precedence:** `canceled = "true"` always takes precedence over `paused` (even if `paused` contains an invalid value). This allows operators to quickly stop runaway processes without needing to fix annotation typos first.

### Execution vs. Observation
* **Execution:** Suppressed while either annotation is active, blocking new instructions from firing.
* **Observation:** Remains active. Probes continue running and status fields persist during pause/cancellation. This ensures `MachineHealthCheck` receives accurate, up-to-date node health data even when a plan is held.

### Pause vs. Cancel Comparison

| Feature Aspect | Pause | Cancel |
| :--- | :--- | :--- |
| **Termination Latency** | Natural (waits for current instruction to finish) | Immediate (`SIGTERM` immediately, `SIGKILL` after 10s) |
| **Process Handling** | Lets process finish naturally; writes checkpoint | Signals entire process group (Unix) or Job Object (Windows) |
| **Resumability** | Yes (checkpoint preserves progress) | No (terminal; requires resetting `plan-state` to `pending`) |
| **Annotation Removal** | Unpauses and resumes execution from checkpoint | Does not un-cancel; requires orchestrator state reset |

## Backward Compatibility (Checksum Flow)

For legacy orchestrators operating without `plan-state` support:
* Annotations generate warning logs but trigger no operational changes.
* Plan reconciliation proceeds normally via checksum comparison.
* Checkpoints, progress persistence, and interrupt watchers remain inactive.

### Unit Tests 
* `interrupt_test.go`: Annotation parsing, interrupt state resolution, precedence rules, polling logic, and edge cases.
* `plan_progress_test.go`: Checkpoint parsing/marshaling, resume state resolution, and sanitization.
* `suppression_test.go`: Matrix of annotation combinations and their impacts on reconciliation.
* `reconcile_test.go` (expanded): `plan-state` transitions, checkpoint updates, and interrupt handling during application.
* `applyinator_test.go` (expanded): Interruption channels, resume-from behaviors, and probe execution while on hold.
* `process_unix_test.go`: Process group termination, signal handling, and zombie reaping.
* `process_windows_test.go`: Job Object termination boundaries and edge cases.

### Integration & E2E Tests  
* `cancellation_test.go`: Full-lifecycle cancellation workflows.
* `pause_test.go`: Pause, resume, and edge-case execution sequences.
* `helpers_test.go`: Framework helpers for annotation manipulation.

Note that E2E Tests for Windows will be added in the rancher/wins repo. See https://github.com/rancher/wins/pull/565 

## Known Limitations

1. **Local Mode Unsupported (`pkg/localplan`):** Pause and cancel capabilities require the remote mode `plan-state` flow. Local plans remain checksum-only without annotation support.
2. **Process Tree Boundaries:** Unix child processes calling `setsid()` escape process group signals and confirmation checks (cgroups are required to fully bridge this gap). A `TerminationIncomplete` flag is populated when this condition is detected to alert operators.

---

## Verification & Testing Instructions

Run the full test suite:
```bash
make test      # Unit tests
make test-e2e  # E2E tests (requires Kind cluster)

### CODE REVIEW

File: pkg/applyinator/applyinator.go [hunk 1]
```diff
@@ -613,9 +840,21 @@ func (a *Applyinator) execute(ctx context.Context, prefix, executionDir string,
 	})
 
 	if err := cmd.Start(); err != nil {
```
jferrazbr: Not sure if it's worth it but should a cancellation check be added here? It might be an edge case, but if cancellation happens during a slow image staging, I think `Start` could still run

jiaqiluo: Hi @jferrazbr, this is a good point. Technically, it is not a problem because the instruction never gets to run meaningfully, and the plan still reports InterruptionCanceled via the resume-checkpoint failure path, but it does waste a process spawn/signal/kill cycle, and momentarily the node runs a command (run.sh or the instruction binary) that has zero chance of completing. Depending on what that binary does in its first fraction of a second (e.g., writes a lockfile, forks children before the process group is even fully set up), that's a small window with side effects that arguably shouldn't have started at all.

That being said, the latest commit adds a check for the ctx before running the image staging or executing the command to prevent them from running if cancellation was invoked.

jferrazbr (dismissed):

pkg/k8splan/reconcile.go
```diff
@@ -180,3 +351,227 @@ func (w *watcher) reconcileSecret(ctx context.Context, sc corecontrollers.Secret
...
+// done is false when neither annotation is active: normal processing continues, using whatever
+// secret and probeStatuses the caller already has. done is true in every other case, and the
+// caller must immediately return (result, err) — for an invalid value, result is secret and err is
+// non-nil; for a recorded interrupt, result is the freshly written Secret (or secret when
+// writeInterruptOutcome's write-once guard made that write a no-op) and err is nil, matching
+// writeInterruptOutcome's own "abandon silently, do not error" contract.
+//
+// currentPlanState is the plan-state to evaluate the interrupt against; callers at different points
+// in reconcileSecret hold different Secrets and must pass whichever plan-state theirs currently
+// carries, not a value cached from earlier in the reconcile.
```
HarrisonWAffel: This comment is helpful but also very hard to follow IMO. Here's the plain text rewrite I think is easier to understand and an explicit suggestion 

```
// checkAndRecordInterrupt evaluates secret.Annotations for any interrupt signals (paused/cancelled).
// If a plan is marked with a signal, checkAndRecordInterrupt will update the PlanCheckpoint 
// field of the provided plan secret with an appropriate value and return the modified 
// secret object. Additionally a boolean 'done' is returned to indicate to the caller that 
// the plan has completed its execution (either temporarily or permanently). 
//
// The returned done value is false when neither annotation is active, indicating 
// to the caller that the plan should be processed using the returned plan secret
// and probeStatuses. In all other cases, done is returned as 'true', indicating to
// the caller that the plan should not be processed further. If an invalid interrupt
// annotation is set on the plan secret, an error is also returned. 
//
// currentPlanState is the plan-state to evaluate the interrupt against. Callers must provide
// whatever state they are currently working against, as opposed to relying on caches which
// may be out of date.
```

```suggestion
// checkAndRecordInterrupt evaluates secret.Annotations for any interrupt signals (paused/cancelled).
// If a plan is marked with a signal, checkAndRecordInterrupt will update the PlanCheckpoint 
// field of the provided plan secret with an appropriate value and return the modified 
// secret object. Additionally a boolean 'done' is returned to indicate to the caller that 
// the plan has completed its execution (either temporarily or permanently). 
//
// The returned done value is false when neither annotation is active, indicating 
// to the caller that the plan should be processed using the returned plan secret
// and probeStatuses. In all other cases, done is returned as 'true', indicating to
// the caller that the plan should not be processed further. If an invalid interrupt
// annotation is set on the plan secret, an error is also returned. 
//
// currentPlanState is the plan-state to evaluate the interrupt against. Callers must provide
// whatever state they are currently working against, as opposed to relying on caches which
// may be out of date.
```

pkg/k8splan/plan_progress.go
```diff
@@ -0,0 +1,100 @@
...
+// resolveResume determines the state and resume position for a reconcile based on the stored plan
+// state and checkpoint. A plan is considered suspended when either plan-state is paused or a valid
+// checkpoint claims the plan. The checkpoint case preserves resumability across agent restarts and
+// handles cases where an external write changed plan-state without clearing the checkpoint.
+// All other states pass through unchanged with resumeFrom set to 0.
+//
+// Precondition: both interrupt annotations have already been validated as explicitly false. This
+// function only determines how to leave an existing suspension; it does not decide whether the
+// plan may leave one. A plan that remains held, or whose annotation cannot be parsed, never reaches
+// this function.
```
HarrisonWAffel: Non blocking nit

```suggestion
```

It's a bit of a pet peeve, but I think a function description should just describe what the function is responsible for, and not necessarily comment on the larger execution flow that the caller is a part of. IMO comments like this can become outdated.

pkg/k8splan/reconcile.go
```diff
@@ -180,3 +351,227 @@ func (w *watcher) reconcileSecret(ctx context.Context, sc corecontrollers.Secret
...
+
+// resumeCommitUpdates returns the Secret updates needed to release a suspension, or nil when the
+// plan was not suspended. The caller has already established that both interrupt annotations are
+// inactive, so the only remaining question is whether a suspension was recorded: plan-state is
+// paused, or the checkpoint for this plan indicates one.
+//
+// Completed, Total and TerminationIncomplete are preserved as a record of what happened to the plan;
+// only the fields that grant a resume are reset. Paused: false revokes the checkpoint's ability to
+// trigger another resume, and ResumeState is cleared because it has already been consumed into
+// effectiveState.
```
HarrisonWAffel:
```suggestion
// resumeCommitUpdates returns the Secret updates needed to release a suspension, or nil when the
// plan was not suspended. The caller must ensure that both interrupt annotations are
// inactive. 
//
// Completed, Total and TerminationIncomplete are preserved as a record of what happened to the plan;
// only the fields that grant a resume are reset. Paused: false revokes the checkpoint's ability to
// trigger another resume, and ResumeState is cleared because it has already been consumed into
// effectiveState.
```

pkg/k8splan/reconcile.go
```diff
@@ -104,38 +160,153 @@ func (w *watcher) reconcileSecret(ctx context.Context, sc corecontrollers.Secret
 		}
 	}
 
+	if resumeUpdates != nil {
+		if needsApplied {
+			logrus.Infof("[k8splan] the plan with checksum %s is no longer held; resuming into plan-state %q from one-time instruction %d of %d",
```
HarrisonWAffel: This log is just announcing that an older plan which used the checksum flow has transitioned to a plan-state approach, right? If so, I don't think that comment would be accurate for net-new plans which never used the old flow. Could be mistaken in my understanding though

pkg/k8splan/reconcile.go
```diff
@@ -104,38 +160,153 @@ func (w *watcher) reconcileSecret(ctx context.Context, sc corecontrollers.Secret
...
+		} else {
+			maps.Copy(secret.Data, resumeUpdates)
+		}
+	}
+
 	output := selectExistingOutput(secret.Data, wasFailedPlan)
 
 	periodicOutput := secret.Data[AppliedPeriodicOutputKey]
 
+	if effectiveState.IsTerminal() && effectiveState != planapi.PlanStateSucceeded && !needsApplied {
```
HarrisonWAffel: IIUC correct this case can only happen when a plan has failed and should not be retried. If so, could we rename the conditional to make that more clear? 

```suggestion
    terminalFailure := effectiveState.IsTerminal() && effectiveState != planapi.PlanStateSucceeded && !needsApplied
	if terminalFailure {
```

non-blocking suggestion

pkg/k8splan/interrupt.go
```diff
@@ -0,0 +1,400 @@
...
+// startInterruptWatch polls the plan Secret while an apply is in flight and closes the channel
+// corresponding to the first interrupt it observes. The controller cannot handle the annotation
+// change directly because Applyinator.Apply runs synchronously in the OnChange handler and
+// DefaultWorkers is 1, leaving the workqueue worker occupied for the duration of the apply.
+// The informer's indexer is updated independently of the worker, so reads from the cache continue
+// to reflect changes while an apply is running.
+//
+// The caller must defer the returned stop function.
+func (w *watcher) startInterruptWatch(ctx context.Context, sc corecontrollers.SecretController) (cancelCh, pauseCh <-chan struct{}, stop func()) {
+	cancel := make(chan struct{})
```
HarrisonWAffel: Does this channel ever get closed on a successful plan execution? I know the interrupt watch explicitly stops the stopChannel, but what about the cancel channel?

HarrisonWAffel (changes requested):
I have some nits and questions but nothing massive. Overall I think this is a solid solution to a very complicated problem, so I think it's important to refine the code comments as much as possible since future people + agents are going to rely on them. 

LGTM from the windows perspective, using Jobs makes sense. I'll take a look at the wins PR next to get a better understanding of how we're testing it all

pkg/k8splan/interrupt.go
jiaqiluo: the cancel channel is not closed, and better not to, because a closed channel would look like a real interrupt if anyone accidentally observed it later. Keeping cancelCh open has no harmful side effect by itself; it will be garbage collected later.

pkg/k8splan/reconcile.go
jiaqiluo: The message is a bit misleading; it is not about handling the transition of an old plan from checksum flow to plan-state flow ( and the transition isn't supported AFAIK), but about handling a paused plan-state plan by clearing that pause and continuing from the stored resume state/checkpoint. 

`if resumeUpdates != nil` gates the logic to plan-state plans. 

A plan can have multiple one-time instructions and can be paused in the middle of executing the n-th instruction, or just finished executing all instructions.  
`needsApplied` will be false if the plan finished all one-time instructions before being paused, so it will not run any one-time instruction after being unpaused. 


Please check the updated messages.

